PR #51, opened 2026-04-27
feat: StreamOutput trait + text/JSON/JSONL formatters + factory (#8)

### LINKED ISSUE #8: A4: Stream formatters — text + json
## Context
`src/stream/output.rs`: Define `StreamOutput` trait with text and json format implementations.

**Design ref:** `docs/designs/2026-04-25-carv-design.md`, Streaming Output section.

## Scope
- `StreamOutput` trait: methods for text, tool_use, tool_result, thinking, done events
- Text formatter: plain text content, tool calls as `[tool: name]` lines
- JSON formatter: single JSON object with full conversation after completion

**Test:** Unit: formatters produce valid output for each event type

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
- **`StreamOutput` trait** — two methods (emit, finish) with boxed futures for object-safety (needed by `Box<dyn StreamOutput>` factory dispatch)
- **`TextFormatter`** — plain-text output, tool calls gated on `--verbose`, usage summary on Done
- **`JsonFormatter`** — accumulates all events, writes `{"events": [...]}` object on finish
- **`StreamJsonFormatter`** — JSONL (one JSON object per line), flushes each event for real-time streaming
- **`create_formatter(format, verbose)`** — factory returning `Box<dyn StreamOutput>`

## Changes
| File | Change |
|------|--------|
| `src/stream/output.rs` | New (529 lines) — event types, trait, 3 formatters, factory, 12 tests |
| `src/stream/mod.rs` | Updated — `pub mod output;` + re-exports |
| `Cargo.toml` | Added `io-std` feature to existing `tokio` dep (no new crates) |

## Verification
- ✅ `cargo build` — 0 errors
- ✅ `cargo test` — 31/31 pass (12 new + 19 existing)
- ✅ `cargo clippy -- -D warnings` — 0 warnings
- ✅ `cargo fmt -- --check` — clean
- ✅ Reviewed by rust-reviewer — approved

Closes #8

### DIFF
--- a/src/stream/output.rs
+++ b/src/stream/output.rs
@@ -102,11 +102,13 @@ impl<W: AsyncWrite + Unpin + Send> StreamOutput for TextFormatter<W> {
             match event {
                 StreamEvent::Text { content } => {
                     self.writer.write_all(content.as_bytes()).await?;
+                    self.writer.flush().await?;
                 }
                 StreamEvent::ToolUse { name, .. } => {
                     if self.verbose {
                         let line = format!("[tool: {name}]\n");
                         self.writer.write_all(line.as_bytes()).await?;
+                        self.writer.flush().await?;
                     }
                 }
                 StreamEvent::Done { turns, usage } => {
@@ -115,6 +117,7 @@ impl<W: AsyncWrite + Unpin + Send> StreamOutput for TextFormatter<W> {
                         usage.input_tokens, usage.output_tokens, usage.cache_read_tokens,
                     );
                     self.writer.write_all(line.as_bytes()).await?;
+                    self.writer.flush().await?;
                 }
                 // Thinking and ToolResult are invisible in text mode.
                 StreamEvent::Thinking { .. } | StreamEvent::ToolResult { .. } => {}
@@ -194,9 +197,9 @@ impl<W: AsyncWrite + Unpin + Send> StreamOutput for StreamJsonFormatter<W> {
         event: StreamEvent,
     ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + '_>> {
         Box::pin(async move {
-            let line = serde_json::to_string(&event)?;
+            let mut line = serde_json::to_string(&event)?;
+            line.push('\n');
             self.writer.write_all(line.as_bytes()).await?;
-            self.writer.write_all(b"\n").await?;
             self.writer.flush().await?;
             Ok(())
         })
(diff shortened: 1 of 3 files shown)


src/stream/output.rs
```diff
@@ -0,0 +1,521 @@
...
+// ---------------------------------------------------------------------------
+// TextFormatter
+// ---------------------------------------------------------------------------
+
+/// Plain-text output formatter.
+///
+/// | Event         | Behaviour                                  |
+/// |---------------|--------------------------------------------|
+/// | `Text`        | Write content directly (no added newline). |
+/// | `Thinking`    | Discarded.                                 |
```
decko, line 79: Why we're discarding Thinking and ToolResults events?

decko (commented):
## Rust Specialist Review — Critical Assessment

A specialist agent reviewed this PR and produced 12 findings. I verified each against the actual code and codebase context. Here's what holds up, what's overstated, and what's noise.

---

### Fix before merge

**`TextFormatter` never flushes between events**
`StreamJsonFormatter` flushes after every `emit()`, but `TextFormatter` does not. For `tokio::io::stdout()`, writes go into an internal buffer and hand off to a blocking thread — tokens will arrive in unpredictable bursts, not character-by-character as users expect from a streaming CLI. This directly undermines the purpose of the text streaming formatter.

Fix: add `self.writer.flush().await?;` at the end of each writing arm in `TextFormatter::emit()`.

**`create_formatter` hardcoded to stdout**
The factory is the only public constructor for `Box<dyn StreamOutput>`, and it's locked to `stdout`. Any integration test of the agent loop will have to capture raw stdout or bypass the factory entirely.

Fix: add a `create_formatter_with_writer(format, verbose, writer)` variant. Keep the convenience `create_formatter` that defaults to stdout.

---

### Nice to fix (trivial)

**Two `write_all` calls per JSONL event**
In `StreamJsonFormatter::emit()`, the line content and newline are written in separate `write_all` calls, each potentially dispatching to the blocking thread. Append `'\n'` to the string before writing:
```rust
let mut line = serde_json::to_string(&event)?;
line.push('\n');
self.writer.write_all(line.as_bytes()).await?;
self.writer.flush().await?;
```

**Missing test for zero-event `JsonFormatter`**
No test covers `JsonFormatter::finish()` when no events have been emitted (should produce `{"events":[]}`). Quick edge case to add.

---

### Correct but overstated (follow-up material, not blockers)

**`JsonFormatter` unbounded memory — LOW, not HIGH.** Its entire purpose is batch accumulation. Users picking `--output-format json` are explicitly asking for this. A documentation note is sufficient.

**`u32` for token counts — LOW, not MEDIUM.** These are per-request counts, not cumulative. A single request won't produce 4B tokens. Defensive `u64` is fine for a follow-up.

---

### Disagree / skip

- **No `Deserialize` on `StreamEvent`** — YAGNI. No consumer needs it yet. One-line follow-up when needed.
- **`emit` should take `&StreamEvent` instead of by value** — Wrong direction. The agent loop produces an event and hands it off; by-value is the natural ownership transfer. Changing to `&StreamEvent` would force `JsonFormatter` to clone on every call, making the common case worse.
- **`ToolResult::content` should be `Value`** — Speculative. The internal event model isn't the API wire format.
- **No `Error` event variant** — Future design question with implications (`should emit continue after errors?`) that shouldn't be solved in this formatter PR.
- **`finish(&mut self)` prevents consuming finalization** — `&mut self` is correct for object-safe traits behind `Box<dyn StreamOutput>`.
- **`verbose` flag threaded to all formats** — Harmless forward-compatibility, not a smell.

---

### Summary

| Priority | Finding | Action |
|---|---|---|
| **Blocker** | `TextFormatter` doesn't flush on `emit()` | Fix — streaming UX bug |
| **Blocker** | Factory hardcoded to stdout | Fix — testability debt |
| Trivial | Two writes per JSONL line | Nice to fix |
| Trivial | Missing zero-event test | Nice to add |
| Follow-up | `u32` → `u64` for usage fields | Low priority |
| Follow-up | Document `JsonFormatter` memory behavior | Low priority |

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

decko:
## Follow-up Review — Post-fix Verification

Commit `941b198` addresses all four items from the previous review. Verified each:

| Finding | Status |
|---|---|
| `TextFormatter` flush on `emit()` | **Fixed** — `flush().await?` added after every `write_all` in all three writing arms (Text, ToolUse, Done) |
| Factory hardcoded to stdout | **Fixed** — `create_formatter_with_writer<W>` added with `'static` bound (correct for boxing), `create_formatter` delegates to it |
| Two `write_all` calls per JSONL line | **Fixed** — `line.push('\n')` before single `write_all` |
| Missing zero-event test | **Fixed** — `json_formatter_zero_events_produces_empty_array` asserts `{"events":[]}` |

No new issues introduced by the fixes. No remaining blockers found.

One minor observation (not blocking): `JsonFormatter::finish()` is not idempotent — calling it twice writes the events array twice (since `self.events` isn't drained). The other two formatters' `finish()` methods are idempotent (just flush). Not a problem given the single-call contract, but worth knowing if the trait contract ever changes.

LGTM — ready to merge.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Issue #22, opened 2026-04-27
D3: OpenAI full stream_chat wiring


decko:
## Context
`src/llm/openai.rs`: Wire together all OpenAI components into a working `LlmProvider` impl.

## Scope
- `OpenAiProvider` struct implementing `LlmProvider`
- Full `stream_chat` method: build request → send → parse SSE stream → emit events

**Type:** 🟢 Formulaic (~80 lines)

**Test:** Integration: mock HTTP server → full streaming response → correct events

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

Issue #21, opened 2026-04-27
D2: OpenAI SSE streaming + reasoning tokens + retry


decko:
## Context
`src/llm/openai.rs`: SSE streaming parser for OpenAI's chat completions. Handle reasoning tokens and retry logic.

## Scope
- SSE stream parsing: `choices[0].delta` events
- `reasoning_content` → `StreamEvent::Thinking`
- Tool call accumulation: `function.arguments` delta fragments
- Retry logic: 3 attempts with exponential backoff

**Type:** 🟠 Dense (~100 lines)

**Test:** Unit: mock SSE chunks → correct StreamEvent sequence with reasoning + tool calls

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

PR #71, opened 2026-05-03
feat(llm): OpenAI SSE streaming provider with reasoning tokens + retry (#21)

### LINKED ISSUE #20: D1: OpenAI wire types + tool schema mapping
## Context
`src/llm/openai.rs`: Serde types for OpenAI's /v1/chat/completions API + ToolDef→OpenAI schema conversion.

## Scope
- Serde types for OpenAI request/response wire format
- Tool schema mapping: provider-agnostic `ToolDef` → OpenAI `{"type": "function", "function": {"name", "description", "parameters"}}`

**Test:** Unit: ToolDef → OpenAI function parameters round-trip

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
- OpenAI SSE streaming provider implementing `LlmProvider::stream_chat`
- `stream::unfold` state machine mapping SSE chunks → `LlmEvent`
- `reasoning_content` → `LlmEvent::Thinking` for o-series models
- Tool call delta accumulation by index with `pending_completions` queue
- Retry logic with exponential backoff (reuses shared `llm::retry` module)
- 9 unit tests covering reasoning content, tool calls, finish_reason variants, message conversion, and usage mapping

## What's Included
1. **`llm::retry`** — extracted shared retry helpers (`NoRetry`, `is_retryable`, `backoff_duration`) from anthropic into a crate-private module. Anthropic updated to import from retry.
2. **`OpenAIProvider`** — struct + `LlmProvider` impl with full SSE state machine, mirroring the Anthropic pattern.
3. **`reasoning_content`** — added to `OpenAIDelta` for o-series model streaming.
4. **`messages_to_openai()`** — converts provider-agnostic `Message` to OpenAI's wire format.

## Depends On
- ✅ #20 (OpenAI wire types) — cherry-picked into this branch. Can close #20 if this is merged first.

## Known Gaps
- `messages_to_openai` does not yet handle tool result messages (`role: "tool"`) or assistant messages with tool calls. Tracked in #22.

## Verification
```
cargo build     ✅
cargo test      ✅ 132 passed
cargo clippy    ✅ 0 warnings
cargo fmt       ✅ clean
```

### DIFF
--- a/src/llm/anthropic.rs
+++ b/src/llm/anthropic.rs
@@ -1,11 +1,11 @@
 //! Anthropic `/v1/messages` API — request (Serialize), SSE (Deserialize), tool mapping.
 use std::collections::HashMap;
-use std::time::Duration;
 
 use futures::{stream, StreamExt};
-use reqwest_eventsource::{retry::RetryPolicy, Event, EventSource};
+use reqwest_eventsource::{Event, EventSource};
 
 use crate::llm::provider::LlmProvider;
+use crate::llm::retry::{backoff_duration, is_retryable, NoRetry, MAX_RETRIES};
 use crate::llm::types::{
     CacheControl, ContentBlock, ContentType, LlmEvent, LlmStream, LlmStreamFuture, LlmUsage,
     Message, RequestConfig, Role, ToolDef,
@@ -115,60 +115,6 @@ pub fn to_anthropic_tools(tools: &[ToolDef]) -> Vec<serde_json::Value> {
             })
         })
         .collect()
-}
-
-// ---------------------------------------------------------------------------
-// Retry policy + backoff helpers
-// ---------------------------------------------------------------------------
-
-/// Retry policy that never retries at the SSE transport layer.
-///
-/// Retry for transient HTTP errors (429, 529) is handled inside the
-/// `stream::unfold` closure by rebuilding the `EventSource` from stored
-/// request state. Letting `EventSource` auto-reconnect would create
-/// duplicate requests and bypass our exponential backoff timing.
-struct NoRetry;
-
-impl RetryPolicy for NoRetry {
-    fn retry(
-        &self,
-        _error: &reqwest_eventsource::Error,
-        _last: Option<(usize, Duration)>,
-    ) -> Option<Duration> {
-        None
-    }
-
-    fn set_reconnection_time(&mut self, _duration: Duration) {}
-}
-
-/// Maximum number of retry attempts for transient HTTP errors (429, 529).
-const MAX_RETRIES: u32 = 3;
-
-/// Backoff base duration — first retry after 1s, then 2s, then 4s.
-const BACKOFF_BASE_MS: u64 = 1000;
-
-/// Check whether an SSE transport error is retryable.
-///
-/// Retryable errors are HTTP 429 (rate limited) and 529 (overloaded).
-/// All other transport errors (connection refused, DNS failure, TLS) are
-/// treated as terminal — they won't resolve within a short backoff window.
-///
-/// Uses `Debug` formatting to inspect the error representation, since
-/// `reqwest_eventsource::Error` does not expose HTTP status codes
-/// through its public API.
-fn is_retryable(error: &impl std::fmt::Debug) -> bool {
-    let msg = format!("{error:?}");
-    msg.contains("429") || msg.contains("529")
-}
-
-/// Compute the backoff duration for the given retry attempt.
-///
-/// Exponential: base * 2^attempt.  Attempt 0 → 1s, attempt 1 → 2s,
-/// attempt 2 → 4s.  Capped at 32s (unreachable with MAX_RETRIES = 3,
-/// but serves as a safety net if MAX_RETRIES is increased).
-fn backoff_duration(attempt: u32) -> Duration {
-    let ms = BACKOFF_BASE_MS * 2u64.pow(attempt.min(5));
-    Duration::from_millis(ms)
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


decko:
## Rust Review — OpenAI SSE streaming provider + retry extraction (#21)

### Summary

This PR adds an OpenAI `/v1/chat/completions` SSE streaming provider that mirrors the existing Anthropic implementation, extracts shared retry helpers into `llm::retry`, and wires up `reasoning_content` for o-series models. The refactoring of `retry` is clean and the `stream::unfold` state machine follows the established pattern well.

---

### Findings

#### Important — Fix Before Merge

**1. `is_retryable` uses `Debug`-string sniffing — fragile and undocumented contract**

`src/llm/retry.rs` lines 1103–1106:
```rust
pub(crate) fn is_retryable(error: &impl std::fmt::Debug) -> bool {
    let msg = format!("{error:?}");
    msg.contains("429") || msg.contains("529")
}
```
The comment acknowledges this, but the tests in `retry.rs` hard-code the precise `Debug` strings (`"InvalidStatusCode(429)"`, `"HTTP error: status code 529"`) that `reqwest_eventsource` currently emits. These are implementation details of a third-party crate and can silently break on a version bump. A false negative means no retry happens and the request fails immediately; a false positive on a string like `"user_id_5299"` would retry a non-transient error.

Recommended fix: inspect `reqwest_eventsource::Error` via `Display` or match on its `reqwest::StatusCode` directly if the public API permits it. If not, at minimum pin the crate version tightly and add a comment marking the tests as regression guards for the specific version in use.

**2. Tool call ordering is not deterministic on `finish_reason = "tool_calls"`**

`src/llm/openai.rs` lines 673–674:
```rust
let completed = std::mem::take(&mut state.tool_calls);
state.pending_completions.extend(completed.into_values());
```
`HashMap::into_values()` yields in arbitrary hash order, not in the original tool-call `index` order. The consumer downstream will see tool calls emitted in a non-deterministic sequence across runs. This is likely harmless today but will cause subtle ordering bugs once multi-turn tool use lands (#22).

Fix: collect `completed` into a `Vec<(u32, ToolAcc)>` first, sort by the key (the `index`), then extend `pending_completions`.

```rust
let mut ordered: Vec<(u32, ToolAcc)> = std::mem::take(&mut state.tool_calls).into_iter().collect();
ordered.sort_unstable_by_key(|(idx, _)| *idx);
state.pending_completions.extend(ordered.into_values());
```

**3. Retry resets partial tool accumulation without telling the consumer**

Lines 551–554:
```rust
state.es = new_es;
state.retry_count += 1;
state.tool_calls.clear();
state.pending_completions.clear();
continue;
```
If a `ToolUseDelta` event was already yielded to the consumer before the retry, clearing `tool_calls` means the final `ToolUseComplete` will never arrive. The consumer will be left with an open partial tool call. The comment in the PR notes this pattern mirrors Anthropic; that same issue exists there too. For now this is acceptable given the `MAX_RETRIES = 3` guard and the fact that retries only occur on 429/529 (not mid-stream), but it should be documented on `OpenAIStreamState` as a known limitation.

---

#### Suggestions — Worth Discussing

**4. `#[allow(dead_code)]` on public wire types**

`OpenAIRequest`, `OpenAIMessage`, `OpenAIToolCall`, `OpenAIFunctionCall`, `OpenAISseChunk`, `OpenAIChoice`, `OpenAIDelta`, `OpenAIToolCallDelta`, `OpenAIFunctionDelta`, `OpenAIUsage` all carry `#[allow(dead_code)]`. These types are constructed and consumed within this file, so the lint is suppressed rather than addressed. Prefer making fields `pub(crate)` where direct external access isn't needed, or drive construction through typed constructors, to recover the dead-code signal.

**5. `messages_to_openai` silently drops non-text content blocks**

Lines 406–413:
```rust
.filter_map(|b| match &b.content {
    crate::llm::types::ContentType::Text { text } => Some(text.as_str()),
    _ => None,
})
```
`ToolUse`, `ToolResult`, and `Thinking` blocks are silently discarded. A `tracing::warn!` call here would help surface the gap during integration before #22 lands, so it doesn't become a silent correctness hole in the agent loop.

**6. The `OpenAIStreamState.max_retries` field is unnecessary**

`max_retries` is always set to `MAX_RETRIES` at construction and never changed. The module-level constant is already in scope inside `stream::unfold` via the closure's capture. Remove the field; use `MAX_RETRIES` directly in the comparison.

**7. `reasoning_content` and `content` on the same chunk are only partially handled**

Lines 604–623: if a chunk carries both `reasoning_content` and `content`, only `reasoning_content` is emitted and the function returns early. The `content` on that chunk is lost until the next `unfold` iteration, at which point there is no chunk to re-read. In practice OpenAI sends them in separate chunks, but the comment should state this assumption explicitly to guard against future model behaviour changes.

**8. Minor: `BACKOFF_BASE_MS` is exported but not needed outside the module**

`pub(crate) const BACKOFF_BASE_MS: u64 = 1000;` is exported from `retry.rs` but only used in `backoff_duration`, which is also in `retry.rs`. Downgrade its visibility to `const` (module-private) to keep the public surface of the module minimal.

---

#### What Was Done Well

- The extraction of retry helpers into `llm::retry` is a clean, justified refactor. Both providers now share one implementation with one set of tests.
- The `OpenAIStreamState` named-struct approach (with a comment explaining why it isn't an anonymous tuple) is a good ergonomic choice for the `stream::unfold` pattern.
- `pending_completions` draining one item per `unfold` iteration is the correct way to yield multiple `ToolUseComplete` events without holding up the stream.
- The `[DONE]` sentinel is handled before JSON parsing, avoiding a spurious deserialization error on the terminal SSE line.
- `finish_reason = "length"` and `"content_filter"` are mapped to `LlmEvent::Error`, which is the right protocol-layer handling.
- Test coverage is solid for the serialization/deserialization layer and the `reasoning_content` path.

---

### Verdict: Request Changes

Two issues should be addressed before merge:

1. Tool call ordering from `HashMap::into_values()` (issue 2) — low-effort fix, correctness impact grows as multi-turn tool use lands.
2. A brief doc comment on the retry-clears-partial-tool-accumulation limitation (issue 3) — no code change required, just documentation.

The `is_retryable` string-sniffing (issue 1) is the most fragile point in the long run but is inherited from the Anthropic provider; fixing both at once in a follow-up is acceptable if the crate version is pinned.

_Reviewed by Rust Specialist Agent_

decko:
## Follow-up Review — 2026-05-03

One new commit has been pushed since the original review: **`b9bdc2d`** — _"fix(llm): address PR #71 review feedback — tool ordering, retry docs, cleanup (#21)"_

---

### Status of Previously Flagged Issues

**Issue 1 — `is_retryable` Debug-string sniffing (Important / flagged as "most fragile long-run")**

Still present and unchanged in `src/llm/retry.rs`. The implementation remains:

```rust
pub(crate) fn is_retryable(error: &impl std::fmt::Debug) -> bool {
    let msg = format!("{error:?}");
    msg.contains("429") || msg.contains("529")
}
```

The original review explicitly acknowledged this was acceptable as a follow-up (to fix both providers at once), and the commit message for `b9bdc2d` does not claim to address it. The only change to `retry.rs` in that commit is the single-character visibility downgrade of `BACKOFF_BASE_MS` from `pub(crate)` to module-private — which addresses suggestion 8 from the original review. The string-sniffing issue and its test suite remain as-is. **This is the agreed-upon disposition: still open, deferred to a follow-up that touches both providers.**

**Issue 2 — `HashMap::into_values()` ordering bug (Important / "fix before merge")**

**Resolved.** `src/llm/openai.rs` lines 549–554 now read:

```rust
let mut ordered: Vec<(u32, ToolAcc)> =
    std::mem::take(&mut state.tool_calls).into_iter().collect();
ordered.sort_unstable_by_key(|(idx, _)| *idx);
state
    .pending_completions
    .extend(ordered.into_iter().map(|(_, acc)| acc));
```

This matches the recommended fix exactly. The inline comment also explains the rationale ("HashMap iteration order is non-deterministic"), which is a nice addition.

---

### Other Feedback Items from Original Review

All remaining items have also been addressed in `b9bdc2d`:

- **Issue 3** (retry-clears-partial-tool limitation undocumented): The `OpenAIStreamState` doc-comment now carries the full limitation description. Resolved.
- **Suggestion 5** (`messages_to_openai` silently drops non-text blocks): A `tracing::warn!` with `content_type` and a `#22` reference is now emitted. Resolved.
- **Suggestion 6** (`max_retries` field unnecessary): Field removed; `MAX_RETRIES` constant is used directly in the comparison. Resolved.
- **Suggestion 7** (`reasoning_content`+`content` same-chunk assumption undocumented): Comment added at the return-early site noting the current API sends them in separate chunks. Resolved.
- **Suggestion 8** (`BACKOFF_BASE_MS` visibility): Downgraded to module-private `const`. Resolved.
- **Suggestion 4** (`#[allow(dead_code)]` on wire types): Still present on all ten wire structs/fields. This was a suggestion, not a required fix, and it was not addressed — acceptable.

---

### Summary

All **Important** issues are resolved (issue 2 fixed; issue 1 deferred by agreement). All **Suggestions** except the `#[allow(dead_code)]` cleanup are resolved. No new issues were introduced by `b9bdc2d`. The one open item is the `is_retryable` Debug-string sniffing, which should be tracked against the follow-up issue for both providers.

**Verdict: Ready to merge.** The open `is_retryable` item does not block this PR given the explicit deferral documented in the original review.

_Follow-up by Rust Specialist Agent_

Issue #68, opened 2026-05-01
Integration: AnthropicProvider HTTP-level SSE integration test


decko:
## Context
The AnthropicProvider's `stream_chat` method (issue #19) has comprehensive unit tests covering SSE event deserialization, retry logic, backoff timing, and cache control annotations. What's missing is an HTTP-level integration test that exercises the full pipeline from HTTP request through SSE stream to `LlmEvent` output.

## Scope
- Add a mock HTTP server (options: `wiremock` as dev-dep, `tokio::net::TcpListener` with `net` feature, or manual)
- Make `ANTHROPIC_BASE_URL` configurable for test injection
- Write test: start mock server → `stream_chat` → verify emitted `LlmEvent` sequence matches expected
- Stretch: test retry path (mock returns 429, then 200)

## Decisions needed
- Mock server approach: `wiremock` (new dev-dep) vs `tokio::net` (new feature flag) vs hand-rolled
- URL override: constructor parameter vs env var vs `#[cfg(test)]` conditional compile

## Type
🟠 Dense (~80 lines)

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

PR #72, opened 2026-05-05
feat(llm): add tool message support to OpenAI provider + fix tool-call completion ordering (#22)

decko:
## Summary
- Adds `Role::Tool` variant and `Message::tool_result()` constructor for tool result messages
- Rewrites `messages_to_openai()` to handle `ToolResult` content blocks (→ `role: "tool"` + `tool_call_id`) and `ToolUse` blocks (→ `tool_calls` array on assistant messages)
- Maps `Role::Tool` → `Role::User` in Anthropic provider (Anthropic has no "tool" role)
- Fixes SSE state machine bug: `ToolUseComplete` events were queued but dropped when `[DONE]` arrived before the queue was drained. Now emits first completion immediately.
- Adds 7 unit tests for tool message conversion + 2 async integration tests with mock TCP server (text streaming + tool call delta accumulation)

## Files changed
- `src/llm/types.rs` (+47): `Role::Tool`, `Message::tool_result()`, serialization test
- `src/llm/anthropic.rs` (+9): Role::Tool → User conversion
- `src/llm/openai.rs` (+360): `messages_to_openai` rewrite, `with_base_url`, state machine fix, tests

## Verification
- [x] `cargo build` ✅
- [x] `cargo test` (140 tests) ✅
- [x] No new dependencies ✅
- [x] DoD review approved ✅

### DIFF
--- a/src/llm/anthropic.rs
+++ b/src/llm/anthropic.rs
@@ -219,6 +219,19 @@ fn extract_system(mut messages: Vec<Message>) -> (Option<AnthropicSystem>, Vec<M
     }
 }
 
+/// Map carv's provider-agnostic roles to Anthropic-compatible roles.
+///
+/// Anthropic does not have a native `role: "tool"` — tool results are sent
+/// as `role: "user"` messages with `tool_result` content blocks.  Call this
+/// before serializing messages to the Anthropic wire format.
+fn normalize_roles_for_anthropic(messages: &mut [Message]) {
+    for msg in messages {
+        if msg.role == Role::Tool {
+            msg.role = Role::User;
+        }
+    }
+}
+
 // ---------------------------------------------------------------------------
 // LlmProvider implementation
 // ---------------------------------------------------------------------------
@@ -233,7 +246,8 @@ impl LlmProvider for AnthropicProvider {
         let api_key = self.api_key.clone();
         let model = self.model.clone();
         let client = self.client.clone();
-        let messages = messages.to_vec();
+        let mut messages = messages.to_vec();
+        normalize_roles_for_anthropic(&mut messages);
         let tools = tools.to_vec();
         let config = config.clone();
 
@@ -453,7 +467,10 @@ impl LlmProvider for AnthropicProvider {
                                     if let Some(BlockState::ToolUse { id, name, json_acc }) =
                                         state.blocks.remove(&index)
                                     {
-                                        match serde_json::from_str(&json_acc) {
+                                        // Default to "{}" for zero-argument tools.
+                                        let json_str =
+                                            if json_acc.is_empty() { "{}" } else { &json_acc };
+                                        match serde_json::from_str(json_str) {
                                             Ok(input) => {
                                                 return Some((
                                                     Ok(LlmEvent::ToolUseComplete {
--- a/src/llm/openai.rs
+++ b/src/llm/openai.rs
@@ -54,6 +54,10 @@ pub struct OpenAIMessage {
     pub content: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tool_calls: Option<Vec<OpenAIToolCall>>,
+    /// Required for `role: "tool"` messages — links the result to the
+    /// assistant's tool call by ID.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tool_call_id: Option<String>,
 }
 
 /// A tool call from the assistant.
@@ -216,6 +220,7 @@ struct OpenAIStreamState {
     done: bool,
     /// Cached request info for rebuilding the EventSource on retry.
     api_key: String,
+    base_url: String,
     request: OpenAIRequest,
     retry_count: u32,
 }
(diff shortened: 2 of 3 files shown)


decko:
## Rust Specialist Code Review — PR #72

### Overview

This PR completes multi-turn tool-use support for the OpenAI provider. The three deliverables are clearly scoped: a `Role::Tool` variant + `Message::tool_result()` constructor, a rewrite of `messages_to_openai()` to handle both `ToolResult` and `ToolUse` blocks, and a fix for the `ToolUseComplete` event drop when `[DONE]` arrives before the pending-completions queue is drained. Verification checklist passes cleanly.

---

### What works well

- **`with_base_url` / mock TCP server pattern** is the correct approach for integration-testing streaming SSE without reaching the network. Avoids mock frameworks and tests real HTTP parsing.
- **Serialization test on `tool_call_id` absence** (`assert!(json.get("tool_calls").is_none())`) correctly distinguishes `null` from missing-key — this matters for OpenAI's strict wire format.
- **SSE bug fix** is correct: returning the first `ToolUseComplete` event directly from the `tool_calls` branch prevents it being queued and dropped when `[DONE]` is processed before the queue is drained.
- **Test for `Role::Tool` serialization round-trip** in `types.rs` gives confidence the serde mapping won't silently regress.

---

### Issues

#### 1. `with_base_url` should be `pub(crate)` or `#[cfg(test)]`

```rust
// current
#[doc(hidden)]
pub fn with_base_url(api_key: String, model: String, base_url: String) -> Self {
```

`#[doc(hidden)]` hides the method from docs but still exports it from the crate. Any downstream user of `carv` as a library can call it. Since this exists solely for testing, the correct visibility is either `pub(crate)` (if the integration tests live in the same crate, which they do) or conditionally compiled:

```rust
#[cfg(test)]
pub(crate) fn with_base_url(api_key: String, model: String, base_url: String) -> Self {
```

#### 2. Empty `json_acc` produces an error, not an empty object

In the SSE fix, when a tool call has no arguments (e.g., a zero-parameter function), `first.json_acc` will be an empty string. `serde_json::from_str("")` returns `Err`, so the provider emits an `LlmEvent::Error` instead of `LlmEvent::ToolUseComplete`. OpenAI always sends at least `{}`, but this is a silent correctness trap for the future:

```rust
// fragile
return match serde_json::from_str(&first.json_acc) {

// defensive
let json_str = if first.json_acc.is_empty() { "{}" } else { &first.json_acc };
return match serde_json::from_str(json_str) {
```

#### 3. Multi-`ToolResult` block data loss is undocumented at the call site

The comment inside `messages_to_openai` correctly notes the last-wins behaviour, but `Message::tool_result()` has no such warning. A future caller constructing a `Message` manually with multiple `ToolResult` blocks (e.g., batching results) would silently lose all but the last. At minimum, add a `tracing::warn!` when more than one `ToolResult` block is detected:

```rust
if m.content.iter().filter(|b| matches!(b.content, ContentType::ToolResult { .. })).count() > 1 {
    tracing::warn!("messages_to_openai: multiple ToolResult blocks in one Role::Tool message; only the last will be sent");
}
```

#### 4. Anthropic role mutation is inline, not a helper

```rust
// anthropic.rs — inside stream_chat
for msg in &mut messages {
    if msg.role == Role::Tool {
        msg.role = Role::User;
    }
}
```

This works, but it's the kind of provider-specific normalization that will be needed in every Anthropic call site (batch, non-streaming, etc.). Extracting it to a small `normalize_roles_for_anthropic(messages: &mut Vec<Message>)` function (even private) keeps the transformation testable and co-located with other Anthropic-specific logic.

---

### Minor nits

- `spawn_sse_server` only accepts a single connection. If a retry path ever exercises that server in a test, `listener.accept()` will hang. Fine for today's tests — worth a one-line comment saying _one-shot_ to prevent future confusion.
- The `collect_events` helper is clean; consider moving it to a shared `tests::helpers` submodule if integration tests grow.

---

### Verdict

**Approve with suggestions.** The core logic is correct, tests are meaningful, and the SSE fix is the right approach. The two actionable items before merge are:

1. Change `with_base_url` to `pub(crate)` (or `#[cfg(test)] pub(crate)`) to avoid leaking a test-only surface.
2. Guard the empty `json_acc` case so zero-argument tools don't emit a spurious `LlmEvent::Error`.

The Anthropic helper extraction and the multi-ToolResult warning are quality improvements but not blockers.

decko:
Addressed all review feedback in 2e02a79:

**Actionable fixes:**
1. **with_base_url visibility** → changed from `#[doc(hidden)] pub` to `pub(crate)` — prevents leaking test-only surface to external consumers
2. **Empty json_acc guard** → all three JSON parse sites (openai pending_completions, openai finish_reason handler, anthropic content_block_stop) now default to `"{}"` for zero-argument tools instead of producing a spurious `LlmEvent::Error`

**Non-blocker improvements:**
3. **Multi-ToolResult warning** → added `tracing::warn!` when more than one ToolResult block is detected in a Role::Tool message (before the fold that keeps only the last)
4. **Anthropic helper extraction** → extracted `normalize_roles_for_anthropic()` function near `extract_system()`, replacing the inline loop in `stream_chat`
5. **One-shot comment** → documented that `spawn_sse_server` accepts only one connection (would hang on retry test patterns)
6. **collect_events relocation note** → added TODO comment suggesting move to shared `tests::helpers` when integration tests grow (e.g., #68)

All 140 tests pass, clippy clean, fmt clean.

PR #75, opened 2026-05-12
feat(tools): add write_file and list_files tools (#24)

### LINKED ISSUE #24: E2: write_file + list_files tools
## Context
`src/tools/fs.rs`: `write_file` and `list_files` tools.

## Scope
- `write_file`: creates/overwrites a file. Invalidates anchor cache + tree-sitter cache on write.
- `list_files`: lists directory contents using `ignore` crate (.gitignore-aware)
- Both implement `Tool` trait with JSON Schema

**Type:** 🟢 Formulaic (~120 lines)

**Test:** Unit: write file → list directory → file appears; write invalidates anchor cache

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
- **WriteFileTool:** creates/overwrites files in the workspace, invalidates anchor cache on write, handles missing params and I/O errors. Guards against path traversal (rejects paths outside workspace_root).
- **ListFilesTool:** lists files (not directories) using `ignore` crate (.gitignore-aware, dot-files excluded), with configurable `max_entries` clamped to 10,000. Errors on non-existent paths and path traversal.
- Both follow the `ReadFileTool` pattern: `Tool` trait, `ToolResult` error handling, workspace-relative path resolution.
- 20 fs tests (162 total)

## Registry wiring
Tool registration (`ToolRegistry::new()`) is **deferred** to the agent loop build-out (future issue). The agent loop (`src/agent/mod.rs`) is currently a stub. All three fs tools are implemented and instantiable — they'll be wired in when the agent loop is built.

## Verification
- [x] `cargo build` — passes
- [x] `cargo test` — 162/162 passes
- [x] `cargo clippy -- -D warnings` — passes
- [x] `cargo fmt -- --check` — passes
- [x] No new dependencies
- [x] No public API changes
- [x] DoD review approved by rust-reviewer

Closes #24

### DIFF
--- a/src/tools/fs.rs
+++ b/src/tools/fs.rs
@@ -1,13 +1,17 @@
-//! `read_file` tool — read file contents with hash-anchored line references.
+//! Filesystem tools — [`ReadFileTool`], [`WriteFileTool`], [`ListFilesTool`].
 //!
-//! Resolves paths relative to the workspace root, canonicalizes when possible,
-//! and returns each line prefixed with its stable anchor word.
+//! These provide the core filesystem operations the LLM agent uses to
+//! read, write, and list files in the project workspace.
 
 use std::path::{Path, PathBuf};
 
 use serde_json::Value;
 
 use crate::tools::traits::{Tool, ToolContext, ToolFuture, ToolResult};
+
+// ---------------------------------------------------------------------------
+// ReadFileTool
+// ---------------------------------------------------------------------------
 
 /// Tool that reads a file and returns its contents with stable anchor identifiers.
 ///
@@ -46,25 +50,19 @@ impl Tool for ReadFileTool {
 
     fn execute<'a>(&'a self, input: Value, ctx: &'a ToolContext) -> ToolFuture<'a> {
         Box::pin(async move {
-            // Extract the "path" parameter.
             let path_str = match input.get("path").and_then(Value::as_str) {
                 Some(p) => p,
                 None => return Ok(ToolResult::error("missing required 'path' parameter")),
             };
 
-            // Resolve relative paths against the workspace root.
             let resolved = if Path::new(path_str).is_absolute() {
                 PathBuf::from(path_str)
             } else {
                 ctx.workspace_root.join(path_str)
             };
 
-            // Canonicalize if possible; fall back to the resolved path on failure
-            // (e.g. the file doesn't exist yet — the I/O error below will be more
-            // informative).
             let canonical = resolved.canonicalize().unwrap_or(resolved);
 
-            // Lock anchor state and read + anchor the file.
             let mut anchor_state = ctx.anchor_state.lock().expect("anchor state lock poisoned");
 
             match anchor_state.get_anchors(&canonical) {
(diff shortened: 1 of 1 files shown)


decko (commented):
## Code Review — Rust Specialist

### Overview

`WriteFileTool` and `ListFilesTool` are well-structured and follow the `ReadFileTool` pattern cleanly. Tests are targeted and the anchor-cache invalidation test is the most valuable one here. A few issues worth addressing before merge:

---

### Correctness Issues

**1. `canonicalize().unwrap_or(resolved)` in `WriteFileTool` creates a cache-coherence hazard**

When writing a **new** file, `canonicalize()` fails (file doesn't exist yet) and falls back to `resolved`. The write and `notify_edit` are both keyed by `resolved`. A subsequent `ReadFileTool` call then calls `canonicalize()` on the now-existing file and gets its true canonical path. If any directory component is a symlink (common on macOS `/tmp → /private/tmp`, Docker, or CI), the invalidation silently becomes a no-op.

```rust
// Current — keyed by pre-write resolved path
let canonical = resolved.canonicalize().unwrap_or(resolved);
match tokio::fs::write(&canonical, content).await {
    Ok(()) => {
        anchor_state.notify_edit(&canonical); // wrong key for new files
```

Fix: canonicalize *after* the write so the key always matches what `ReadFileTool` will use:

```rust
match tokio::fs::write(&resolved, content).await {
    Ok(()) => {
        let written_canonical = resolved.canonicalize().unwrap_or(resolved);
        anchor_state.notify_edit(&written_canonical);
```

---

**2. `u64 → usize` cast silently truncates on 32-bit targets**

```rust
.map(|v| v as usize)
```

Use `usize::try_from(v).unwrap_or(200)` instead.

---

### Integration Gap

**3. Neither new tool is registered anywhere**

After grepping every Rust file outside `fs.rs`, `WriteFileTool` and `ListFilesTool` are never instantiated or passed to `ToolRegistry::new` — the LLM agent cannot invoke them. If registry wiring is intentionally deferred to a follow-on PR, that should be called out explicitly (and ideally tracked as a follow-up issue), since without it these tools have zero runtime effect.

---

### Minor

**4. `write_to_subdirectory` tests failure, not success** — rename to `write_fails_when_parent_missing` for clarity.

**5. `let _ = write!(output, ...)` is unnecessary** — writing to a `String` is infallible; `write!` on `String` always returns `Ok(())`. Drop the `let _ =` suppressor.

**6. Test isolation: `temp_dir` directories are not cleaned up** — `list_empty_directory` is the most vulnerable: stale state from a previous run will cause it to return entries instead of `"(empty directory)"`. Consider `tempfile::TempDir` or a unique suffix per run.

**7. `list_respects_ignore_filters` comment hedges unnecessarily** — `standard_filters(true)` does not hide `.gitignore` itself (only hidden *directories*), so the file will consistently appear. Assert it appears rather than leaving it as "may or may not."

---

### Positive Highlights

- `is_read_only() -> false` override on `WriteFileTool` is correct.
- `entries.sort()` ensures deterministic output from the walker.
- `notify_edit` called only after confirmed successful write — correct ordering.
- `require_git(false)` in `WalkBuilder` is the right call for non-git workspaces.
- The anchor cache invalidation test is well-constructed and tests the right invariant.

Issues #1 and #3 are worth fixing before merge; the rest are low-risk improvements.

decko:
Good catches — all addressed in 9d3814d:

1. **Cache-coherence (#1):** ✅ Fixed. WriteFileTool now canonicalizes AFTER the write. For new files, the post-write canonicalize gets the true symlink-resolved path, matching what ReadFileTool will later resolve. Also added an explanatory comment.

2. **u64→usize truncation (#2):** ✅ Fixed. Replaced `as usize` with `usize::try_from(v).unwrap_or(200)`.

3. **Registry wiring (#3):** This is expected at this stage. The agent loop (`src/agent/mod.rs`) is a stub — it's a single comment line. The `ToolRegistry::new()` calls are all in tests. Wiring real tools into the agent loop is a future integration step (the agent loop and its prompt→LLM→tool→repeat logic haven't been built yet). All three tools (ReadFile, WriteFile, ListFiles) are implemented and instantiable — they just need a registry call in the agent loop when it's built.

4. **Test name (#4):** ✅ Renamed `write_to_subdirectory` → `write_fails_when_parent_missing`.

5. **`let _ = write!(...)` (#5):** ✅ Replaced with `.expect("write to String is infallible")`.

6. **Test isolation (#6):** ✅ Fixed. `temp_dir` now runs `remove_dir_all` before `create_dir_all`, giving each test a clean slate regardless of stale state from previous runs.

7. **Comment hedging (#7):** ✅ Fixed — but in the opposite direction. `standard_filters(true)` enables the `hidden` filter, which DOES hide dot-files including `.gitignore` itself. Changed the hedged comment to an explicit assertion that `.gitignore` is absent, with an explanation of why.

decko (commented):
## Re-Review — Follow-up on Previous Feedback

### Resolved ✅

All six specific issues from the first review were addressed:

- **`canonicalize()` cache-coherence bug** — fixed correctly: write to `resolved`, canonicalize after success, then `notify_edit`. The explanatory comment is a nice touch.
- **`u64 → usize` truncation** — fixed with `usize::try_from(v).unwrap_or(200)`.
- **`write_fails_when_parent_missing` rename** — done.
- **`let _ = write!(...)` suppressor** — replaced with `.expect("write to String is infallible")`.
- **Test isolation** — `remove_dir_all` before `create_dir_all` in `temp_dir`; comment documents intent.
- **`.gitignore` visibility in test** — now asserts absence and explains why (`standard_filters` hides dot-files).

---

### Still Unresolved ❌

**Tools not registered — still dead code at runtime.**
`WriteFileTool` and `ListFilesTool` are never instantiated or passed to `ToolRegistry::new` anywhere outside `fs.rs`. The agent cannot invoke them. If registry wiring is deferred to a follow-on PR, the description should say so explicitly and link the tracking issue.

---

### New Observations

**`ListFilesTool` output includes directories but description says "files"**

`WalkBuilder` yields both files and directories. The description says *"List files in a directory … Returns relative paths, one per line"* — nothing tells the LLM that directory entries like `src/` will appear alongside `src/main.rs`. This can cause confusion when the LLM tries to open a returned path. Either filter to files-only:

```rust
if entry.file_type().map_or(false, |ft| ft.is_file()) {
    // push to entries
}
```

…or update the description to say *"files and directories."*

**No upper bound on `max_entries`**

Passing `max_entries: 9999999` against a broad path triggers an unbounded walk. The default 200 is sensible, but there's no ceiling. Consider clamping: `usize::try_from(v).unwrap_or(200).min(10_000)`.

**`standard_filters(true)` hides dot-files but description omits this**

The description mentions `.gitignore` and `.ignore` patterns but not the hidden-file filter that `standard_filters(true)` also activates (`.env`, `.cargo/`, etc. silently disappear). Suggest adding: *"Hidden files and directories (dot-prefixed) are also excluded."*

---

### Minor

- `f.flush()` in `write_temp_file` is a no-op for `std::fs::File` — harmless but unnecessary.
- `read_existing_file` anchor coverage was weakened (single `contains` check vs. the old per-line loop) — not this PR's focus but worth knowing.

The missing registry wiring is the remaining blocker. The description/filter gap and the directory-vs-file ambiguity are quick fixes worth doing before merge.

decko:
All three new observations fixed in 233acc5:

- **Files-only filter:** `ListFilesTool` now skips directory entries. Only file paths are returned.
- **max_entries clamp:** Capped at `.min(10_000)` — prevents unbounded walks.
- **Description updates:** Doc comment and `description()` now mention dot-file filtering (hidden files/directories are excluded by `standard_filters(true)`).

Also removed the unnecessary `f.flush()` on `std::fs::File` and updated the PR description to explicitly note that registry wiring is deferred to the agent loop build-out.

decko (commented):
## Third Review — All Previous Feedback Resolved ✅

Every item from the first two reviews has been addressed. No unresolved blockers remain. Two new findings below.

---

### New Finding 1 — `WriteFileTool` has no workspace boundary check (path traversal)

Both relative and absolute paths are accepted without validating they stay inside `workspace_root`:

```rust
let resolved = if Path::new(path_str).is_absolute() {
    PathBuf::from(path_str)           // any absolute path accepted
} else {
    ctx.workspace_root.join(path_str) // "../../etc/passwd" escapes root
};
```

`ReadFileTool` carries the same pattern (pre-existing), but read access outside the workspace is lower risk than write. An LLM that is tricked or confused could overwrite arbitrary files on the host. A minimal guard closes it:

```rust
let canonical = resolved.canonicalize().unwrap_or(resolved.clone());
if !canonical.starts_with(&ctx.workspace_root) {
    return Ok(ToolResult::error("write_file failed: path escapes workspace root"));
}
```

Whether to fix it here or in a dedicated hardening pass is a judgment call, but it deserves a conscious decision before merge.

---

### New Finding 2 — Stale comment in `list_directory_contents` test

```rust
// Output should contain all files (and subdirectory).
```

Since the files-only filter was added, the `sub` directory entry itself no longer appears in output. The test assertions are still correct (they check file paths only), but the parenthetical `(and subdirectory)` is now wrong and will mislead future readers.

---

Otherwise the PR is in good shape — correct implementation, all prior issues resolved, solid test coverage. The path traversal item is the one thing worth deciding on before merge.

decko:
Both findings fixed in 55192e2:

**1. Workspace boundary check:** `WriteFileTool` now validates the resolved path stays within `workspace_root` before writing. Walks up to the nearest existing ancestor for canonicalization when the target file is new (otherwise canonicalize fails and we can't compare). Absolute paths and `../../` escapes are both caught.

**2. Stale comment:** Removed the `(and subdirectory)` parenthetical since the files-only filter no longer returns directory entries.

decko (commented):
## Fourth Review — All Prior Findings Resolved ✅

Both items from the third review (path traversal guard + stale comment) have been addressed. The guard implementation is correct: walking up to the nearest existing ancestor before canonicalizing handles new files, symlinked parents, and absolute paths properly.

---

### Remaining Item 1 — No test for the path traversal guard

The guard is the most security-sensitive new code in this PR and has no dedicated test. `write_fails_when_parent_missing` covers a different failure path. A single test closes this gap:

```rust
#[tokio::test]
async fn write_rejects_path_outside_workspace() {
    let dir = temp_dir("write_traversal");
    let ctx = test_context(dir);

    let tool = WriteFileTool;
    let result = tool
        .execute(json!({"path": "../../outside.rs", "content": "oops"}), &ctx)
        .await
        .unwrap();

    assert!(result.is_error);
    assert!(result.content.contains("path escapes workspace root"));
}
```

---

### Remaining Item 2 — `ListFilesTool` has no workspace boundary check

The guard was added to `WriteFileTool` only. `ListFilesTool` still accepts paths like `/etc` or `../../sensitive_dir` and will walk them. It's read-only so the risk is lower, but it's asymmetric: an LLM could be directed to enumerate the host filesystem. A `canonical.starts_with(&workspace_root_canon)` check after the existing `is_dir()` guard would close it consistently.

---

No correctness bugs remaining. The missing traversal test is the one item worth adding before merge.

decko:
Both items fixed in d52ac0c:

**1. Traversal test:** Added `write_rejects_path_outside_workspace` — verifies `../../` escape is caught with the right error message.

**2. ListFilesTool boundary check:** Added the same workspace check after the `is_dir()` guard. Now both write and list tools reject paths that escape `workspace_root`.

decko (commented):
## Fifth Review — Both Prior Items Resolved ✅

`write_rejects_path_outside_workspace` test added and `ListFilesTool` workspace boundary guard added. Two small items remain:

---

**1. No test for `ListFilesTool`'s new traversal guard**

`WriteFileTool` got `write_rejects_path_outside_workspace` — `ListFilesTool` got the same guard but no corresponding test. A symmetric test would close the gap:

```rust
#[tokio::test]
async fn list_rejects_path_outside_workspace() {
    let dir = temp_dir("list_traversal");
    let ctx = test_context(dir);

    let tool = ListFilesTool;
    let result = tool
        .execute(json!({"path": "../../"}), &ctx)
        .await
        .unwrap();

    assert!(result.is_error);
    assert!(result.content.contains("path escapes workspace root"));
}
```

**2. PR description test count is stale**

The summary still says "18 fs tests (160 total)" after `write_rejects_path_outside_workspace` was added, making it 19 fs tests. The verification checklist recount and total should be updated accordingly.

---

Otherwise merge-ready.

decko:
Both items fixed in 7647b01:

**1. ListFilesTool traversal test:** Added `list_rejects_path_outside_workspace` — verifies `../../` escape is rejected by the listing tool symmetrically with the write tool.

**2. PR description updated:** Test counts corrected to 20 fs tests / 162 total.

decko (commented):
## Sixth Review — All Prior Items Resolved ✅

`list_rejects_path_outside_workspace` added and PR description counts corrected. The implementation is correct, all guards are tested, test isolation is solid.

---

### One Minor Remaining Item

**`path` parameter descriptions don't mention the workspace boundary**

Both tools' `parameters_schema` describe `path` as "relative to the project root or absolute" — but absolute paths outside `workspace_root` are now rejected. An LLM following the schema description literally might try `/home/user/other_project/file.rs` and receive a confusing `path escapes workspace root` error with no hint from the schema that this would fail.

A small wording change in both schemas:

```rust
// WriteFileTool
"description": "Path to the file to write, relative to the project root or absolute within the workspace"

// ListFilesTool
"description": "Path to the directory to list, relative to the project root or absolute within the workspace (defaults to the project root)"
```

---

That's the only remaining item. Ready to merge once addressed.

decko:
Fixed in 274b830: both WriteFileTool and ListFilesTool `path` schema descriptions now say "absolute within the workspace" instead of just "absolute."

Issue #35, opened 2026-04-27
H1: System prompt construction


decko:
## Context
`src/agent/context.rs`: Build the system prompt that sets up the agent's behavior.

## Scope
- Include: cwd, git branch, available tools list, hash-anchor editing rules
- Available tools section: list all tools from registry with descriptions
- Hash-anchor rules: explain anchor format, disambiguation, edit operations
- Use `--system-prompt` if provided, otherwise default

**Type:** 🟢 Formulaic (~100 lines)

**Test:** Unit: system prompt includes cwd, git branch, tool list, anchor editing rules

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

PR #76, opened 2026-05-14
feat(tools): add search_files tool (ripgrep-based, .gitignore-aware) (#25)

### LINKED ISSUE #25: E3: search_files tool (ripgrep-based, .gitignore-aware)
## Context
`src/tools/search.rs`: Content search using ripgrep engine (`grep-regex` + `grep-searcher`). Returns anchor-prefixed matching lines.

## Scope
- `search_files` tool: accepts pattern and path, searches using ripgrep
- .gitignore-aware via `ignore` crate
- Results returned with hash-anchored lines
- Tool is read-only

**Type:** 🟢 Formulaic (~120 lines)

**Test:** Unit: search fixture directory → returns anchored matches

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
- Implements `SearchFilesTool` — a read-only ripgrep-based content search tool
- `.gitignore`-aware via `ignore::WalkBuilder` with `standard_filters(true)`
- Returns hash-anchored matching lines (`{anchor}│{line}`) — same format as `ReadFileTool`
- 9 tests covering parameter validation, gitignore respect, path scoping, deduplication, and edge cases

## Changes
| File | Lines | Description |
|------|-------|-------------|
| `src/tools/search.rs` | +458 | New `SearchFilesTool` implementation + test suite |
| `src/tools/mod.rs` | +12 | Module declaration + `default_tools()` helper |

## Dependencies
No new dependencies — `ignore`, `grep-regex`, and `grep-searcher` were already in `Cargo.toml`.

## Checks
- [x] `cargo build` ✅
- [x] `cargo test` (158 tests) ✅
- [x] DoD review passed (rust-reviewer)

## Notes
- Line count (470) exceeds 150-line FORMULAIC cap — ~314 lines are tests; tool impl is ~144 lines
- `default_tools()` forward-looking helper for agent loop wiring (#35-38)
- Reviewer flagged minor optimizations (unnecessary `to_path_buf()`, pre-allocation) as non-blocking

Closes #25

### DIFF
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -6,7 +6,19 @@
 
 pub mod fs;
 pub mod registry;
+pub mod search;
 pub mod traits;
 
 pub use registry::ToolRegistry;
 pub use traits::{Tool, ToolContext, ToolFuture, ToolResult};
+
+/// Returns the default set of tools available to the LLM.
+///
+/// The agent loop calls this to populate [`ToolRegistry`] at startup,
+/// filtering out any disallowed tools afterward.
+pub fn default_tools() -> Vec<Box<dyn Tool>> {
+    vec![
+        Box::new(fs::ReadFileTool),
+        Box::new(search::SearchFilesTool),
+    ]
+}
--- a/src/tools/search.rs
+++ b/src/tools/search.rs
@@ -12,6 +12,7 @@ use grep_regex::RegexMatcher;
 use grep_searcher::{sinks, Searcher};
 use ignore::WalkBuilder;
 use serde_json::Value;
+use tracing::debug;
 
 use crate::tools::traits::{Tool, ToolContext, ToolFuture, ToolResult};
 
@@ -45,6 +46,10 @@ impl Tool for SearchFilesTool {
                 "path": {
                     "type": "string",
                     "description": "Directory to search, relative to the project root or absolute (defaults to workspace root)"
+                },
+                "max_results": {
+                    "type": "integer",
+                    "description": "Maximum number of matching lines to return (default 500)"
                 }
             },
             "required": ["pattern"]
@@ -70,6 +75,13 @@ impl Tool for SearchFilesTool {
                 None => ctx.workspace_root.clone(),
             };
 
+            // Extract optional "max_results" parameter; default to 500.
+            let max_results: usize = input
+                .get("max_results")
+                .and_then(|v| v.as_u64())
+                .map(|v| v as usize)
+                .unwrap_or(500);
+
             // Compile the regex pattern.
             let matcher = match RegexMatcher::new(pattern) {
                 Ok(m) => m,
@@ -83,24 +95,28 @@ impl Tool for SearchFilesTool {
 
             let mut searcher = Searcher::new();
             let mut all_results: Vec<String> = Vec::new();
+            let mut truncated: usize = 0;
 
             for entry in walker {
                 let entry = match entry {
                     Ok(e) => e,
-                    Err(_) => continue,
+                    Err(e) => {
+                        debug!("skipping unreadable entry: {e}");
+                        continue;
+                    }
                 };
 
                 if !entry.file_type().is_some_and(|ft| ft.is_file()) {
                     continue;
                 }
 
-                let file_path = entry.path().to_path_buf();
+                let file_path = entry.path();
 
                 // Use grep-searcher to find matching lines.
                 let mut matched_line_nums: BTreeSet<u64> = BTreeSet::new();
                 let search_result = searcher.search_path(
                     &matcher,
-                    &file_path,
+                    file_path,
                     sinks::UTF8(|line_num, _line| {
                         matched_line_nums.insert(line_num);
                         Ok::<_, std::io::Error>(true)
@@ -108,7 +124,8 @@ impl Tool for SearchFilesTool {
                 );
 
                 // Skip files that can't be searched (binary, permission, etc.).
-                if search_result.is_err() {
+                if let Err(e) = search_result {
+                    debug!("skipping unsearchable file {}: {e}", file_path.display());
                     continue;
                 }
 
@@ -116,20 +133,41 @@ impl Tool for SearchFilesTool {
                     continue;
                 }
 
+                // Compute path relative to workspace root for output.
+                let relative_path = file_path
+                    .strip_prefix(&ctx.workspace_root)
+                    .unwrap_or(file_path);
+
                 // Get anchors for this file.
                 let mut anchor_state = ctx.anchor_state.lock().expect("anchor state lock poisoned");
-                let anchors = match anchor_state.get_anchors(&file_path) {
+                let anchors = match anchor_state.get_anchors(file_path) {
                     Ok(a) => a,
-                    Err(_) => continue,
+                    Err(e) => {
+                        debug!(
+                            "skipping file with anchor error {}: {e}",
+                            relative_path.display()
+                        );
+                        continue;
+                    }
                 };
 
                 // Map 1-indexed line numbers from grep-searcher to 0-indexed anchor positions.
                 for line_num in &matched_line_nums {
+                    if all_results.len() >= max_results {
+                        truncated += 1;
+                        continue;
+                    }
                     let idx = line_num.saturating_sub(1) as usize;
                     if let Some((anchor, line)) = anchors.get(idx) {
-                        all_results.push(format!("{anchor}│{line}\n"));
+                        all_results.push(format!("{}:{anchor}│{line}\n", relative_path.display()));
                     }
                 }
+            }
+
+            if truncated > 0 {
+                all_results.push(format!(
+                    "... {truncated} more matches omitted. Narrow your search pattern.\n"
+                ));
             }
 
             if all_results.is_empty() {
@@ -157,6 +195,7 @@ mod tests {
     /// Create (or re-use) a temporary directory inside the OS temp dir.
     fn temp_dir(name: &str) -> PathBuf {
         let dir = std::env::temp_dir().join(name);
+        let _ = std::fs::remove_dir_all(&dir); // clean stale state from prior runs
         std::fs::create_dir_all(&dir).expect("failed to create temp dir");
         dir
     }
@@ -194,6 +233,11 @@ mod tests {
         assert!(
             !result.is_error,
             "expected success, got: {}",
+            result.content
+        );
+        assert!(
+            result.content.contains("test.rs:"),
+            "missing file path prefix in:\n{}",
             result.content
         );
         assert!(
@@ -324,11 +368,17 @@ mod tests {
         // With gitignore working we see 1 unique line; without it we'd see 2.
         let unique_lines: std::collections::HashSet<&str> = result.content.lines().collect();
 
-        assert!(
-            unique_lines.len() <= 1,
-            "expected at most 1 unique line (gitignored file excluded), got {}: {:?}",
+        assert_eq!(
+            unique_lines.len(),
+            1,
+            "expected exactly 1 unique line (gitignored file excluded), got {}: {:?}",
             unique_lines.len(),
             unique_lines
+        );
+        assert!(
+            result.content.contains("foo.txt:"),
+            "expected 'foo.txt:' file path prefix in output, got: {}",
+            result.content
         );
         assert!(
             result.content.contains("│secret stuff"),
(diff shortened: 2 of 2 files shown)


decko:
## Code Review — Rust & Software Engineering

**Summary:** Solid, idiomatic implementation. The tool is well-scoped, the output format aligns with `ReadFileTool`, and test coverage is meaningful. A few real issues need attention before merge.

---

### Correctness Issues

#### 1. Output format omits the file path — the LLM can't know which file to edit

`ReadFileTool` targets a single file, so `{anchor}│{line}` is sufficient. `SearchFilesTool` spans the entire workspace, so the LLM receives matching lines with no indication of origin:

```
WORDWORD│fn hello() {
WORDWORD│    println!("hello");
```

If the same line appears in `src/a.rs` and `src/b.rs`, the LLM cannot know where to apply a fix. The anchor alone isn't enough — it's a hash of the line, not the file. The output should include the file path, e.g.:

```
src/a.rs:WORDWORD│fn hello() {
```

or group matches under a file header (grep-style). This is a functional gap that will surface in any multi-file codebase.

#### 2. `gitignore_respected` test has a logical flaw

```rust
let unique_lines: std::collections::HashSet<&str> = result.content.lines().collect();
assert!(unique_lines.len() <= 1, ...);
```

`<= 1` passes when `unique_lines.len() == 0` — i.e., when there are *no* results at all. If the gitignore handling silently broke *both* files, the test would still pass. The correct assertion is `== 1`: exactly one match (from `foo.txt`), with `bar.txt` excluded. Change to:

```rust
assert_eq!(unique_lines.len(), 1, "expected exactly 1 match (gitignored file excluded), got ...");
```

---

### Bugs / Risks

#### 3. Mutex held across file I/O in `get_anchors`

```rust
let mut anchor_state = ctx.anchor_state.lock().expect("anchor state lock poisoned");
let anchors = match anchor_state.get_anchors(&file_path) { ... };
```

`get_anchors` reads and hashes the file while the lock is held. For every matched file this serializes disk I/O behind the mutex. This is fine today (single-threaded agent loop) but will become a bottleneck the moment the walker is parallelized (which `WalkBuilder` explicitly supports via `.build_parallel()`). Consider caching read anchors in `AnchorState` keyed by `(path, mtime/inode)` so the hot path is a map lookup, not a re-read.

#### 4. Silent error suppression without any observability

```rust
let entry = match entry {
    Ok(e) => e,
    Err(_) => continue,  // walk error swallowed
};
...
let anchors = match anchor_state.get_anchors(&file_path) {
    Ok(a) => a,
    Err(_) => continue,  // anchor error swallowed
};
```

Binary files and permission errors are expected here, so skipping is correct — but with no `tracing::warn!` or similar, there's no way to distinguish "this file was skipped because binary" from "this file was skipped because of a bug." Even a `tracing::debug!` would help.

#### 5. No result cap — unbounded output for broad patterns

A pattern like `.` or `fn ` on a large repo will return every line in every non-ignored file. There is no `max_results` parameter and no truncation. The LLM context window will overflow silently. Suggest a configurable hard cap (e.g., 500 lines) with a trailing note: `"... N more matches omitted. Narrow your pattern."`.

---

### Minor / Non-blocking

#### 6. Unnecessary `to_path_buf()` allocation

```rust
let file_path = entry.path().to_path_buf();   // allocates
searcher.search_path(&matcher, &file_path, ...)
```

`search_path` takes `AsRef<Path>`. `entry.path()` returns `&Path` directly and satisfies that bound. The `to_path_buf()` can be dropped on the `search_path` call site; you still need the owned `PathBuf` for the `get_anchors` call, so keep the binding but use it only where `'static` lifetime matters:

```rust
let file_path = entry.path();   // &Path, no alloc
searcher.search_path(&matcher, file_path, ...)?;
...
anchor_state.get_anchors(file_path)
```

(Noted as non-blocking in the PR already — flagging because it has a straightforward fix.)

#### 7. Test isolation — shared temp directories not cleaned up

`temp_dir(name)` uses `create_dir_all` and never removes its output. Tests reuse the same paths across runs, so a test that fails mid-write can leave stale files that contaminate later runs. The `ReadFileTool` tests have the same pattern, so this is a repo-wide convention to address, but this PR is a good place to start since it adds more of them. At minimum, pre-clean with `let _ = std::fs::remove_dir_all(&dir);` before `create_dir_all`.

---

### Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **Blocking** | `search.rs` | Output missing file path — LLM can't locate matches |
| 2 | **Blocking** | `search.rs` | `gitignore_respected` test assert allows false-pass |
| 3 | Medium | `search.rs` | Mutex held during file I/O — scalability concern |
| 4 | Medium | `search.rs` | Silent error suppression — zero observability |
| 5 | Medium | `search.rs` | No result cap — unbounded LLM context consumption |
| 6 | Low | `search.rs` | Unnecessary `to_path_buf()` allocation |
| 7 | Low | `search.rs` / `fs.rs` | Test temp dirs not cleaned between runs |

Items 1 and 2 need fixes before merge. The rest can be follow-ups if needed.

decko:
All findings addressed in 8caf3e4:

| # | Status | Fix |
|---|--------|-----|
| 1 | ✅ | Output now prefixes each line with relative path: `src/tools/search.rs:WORDWORD│line` |
| 2 | ✅ | `gitignore_respected` assertion changed to `assert_eq!(unique_lines.len(), 1, ...)` |
| 4 | ✅ | `tracing::debug!()` added at all three error-silence points |
| 5 | ✅ | Added optional `max_results` param (default 500) with `"... N more matches omitted"` truncation |
| 6 | ✅ | Removed `to_path_buf()`, using `entry.path()` directly |
| 7 | ✅ | Test `temp_dir()` now calls `remove_dir_all` before `create_dir_all` |

**Item 3** (mutex held during I/O) deferred — matches existing `ReadFileTool` pattern and the agent loop is single-threaded.

All checks: `cargo build` ✅ `cargo test` (158) ✅ `cargo clippy` ✅ `cargo fmt` ✅

decko:
## Follow-up Review — Post-Fix Pass

All 7 items from the first review are resolved. Three new observations on the updated code:

---

### New: Truncation walks the full tree after the cap is hit

```rust
for entry in walker {
    ...
    for line_num in &matched_line_nums {
        if all_results.len() >= max_results {
            truncated += 1;
            continue;           // counts omitted lines …
        }
        // … but the outer loop never breaks
    }
}
```

Once `max_results` is reached, the tool still walks every remaining file in the workspace, compiles anchors for each matched file, and counts individual omitted lines. On a large repo with a broad pattern (`fn`, `let`, `use`) this makes the cap useless as a latency guard — it stops collecting output but doesn't stop the work.

Two viable fixes:

**Option A — break early, report approximate count:**
```rust
if all_results.len() >= max_results {
    // walk just enough to count, no anchor work needed
    break;  // add a note: "and possibly more"
}
```

**Option B — count only at the file level before anchor lookup:**
```rust
if all_results.len() >= max_results {
    truncated += matched_line_nums.len();
    continue;  // skip anchor lock + read entirely
}
```

Option B is better: it avoids the `anchor_state` lock for all post-cap files while still giving an exact count. The inner per-line loop becomes unreachable once the cap is hit, so the `truncated += matched_line_nums.len()` replaces it entirely.

---

### Weakened anchor-presence check in `search_matching_lines`

The original test verified that each output line had a non-empty anchor before `│`. The new format is `path:anchor│line`, so:

```rust
let before = line.split('│').next().unwrap_or("");
assert!(!before.is_empty(), ...);
```

`before` is now `"test.rs:ANCHOR"` — always non-empty as long as there's a path, even if the anchor portion is empty (e.g., `"test.rs:"│line`). The test no longer catches a regression where anchors are dropped. Tighten it:

```rust
let before = line.split('│').next().unwrap_or("");
// before is "path:anchor" — split on ':' and check the last segment
let anchor_part = before.rsplitn(2, ':').next().unwrap_or("");
assert!(!anchor_part.is_empty(), "anchor should not be empty in line: {line:?}");
```

---

### `max_results: 0` silently returns the truncation message instead of matches

With `max_results: 0`:
- `all_results.len() >= max_results` is `0 >= 0` → true on the first line
- All matches land in `truncated`
- `all_results` stays empty until the truncation message is pushed
- Return: `"... N more matches omitted. Narrow your search pattern."` with no actual results

The LLM receives the truncation message but can't narrow the pattern further because it saw nothing. A minimum of 1 should be enforced, or at least validated:

```rust
let max_results: usize = input
    .get("max_results")
    .and_then(|v| v.as_u64())
    .map(|v| (v as usize).max(1))   // floor at 1
    .unwrap_or(500);
```

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| A | Medium | Outer walk continues after cap — anchor work not skipped |
| B | Low | Anchor-presence assertion weakened by new path prefix |
| C | Low | `max_results: 0` returns only a truncation message, no results |

No blocking issues remain. Item A is the most impactful to fix before merge; B and C can be follow-ups.

decko:
Round 2 findings addressed in 7292c87:

| # | Fix |
|---|-----|
| A | Cap check moved **before** anchor lock — post-cap files skip `anchor_state` entirely, count matched lines at file level with `continue` |
| B | Test anchor check now `rsplitn(2, ':')` to isolate anchor from path prefix |
| C | `max_results` extraction floored at `.max(1)` — `0` treated as `1` |

`cargo build` ✅ `cargo test` (158) ✅ `cargo clippy` ✅ `cargo fmt` ✅

decko:
## Round 3 Review — Single Finding

All three items from the last round (A, B, C) are correctly fixed. One regression was introduced in the process.

---

### Cap bypass when a single file has more matches than `max_results`

The fix for issue A moved the cap check to the file level (before the anchor lock) and removed the per-line check inside the inner loop. This is correct for skipping *subsequent files* after the cap, but it leaves a gap: a single file whose match count exceeds `max_results` bypasses the cap entirely.

**Reproduce:** `max_results=2`, one file with 5 matching lines.

```
// file-level check: 0 >= 2 → false → proceed
if all_results.len() >= max_results { ... }

// inner loop — no guard
for line_num in &matched_line_nums {
    // pushes all 5 lines unconditionally
    all_results.push(format!(...));
}
// all_results.len() == 5, max_results == 2
```

The fix is to keep both guards — the file-level check (for the anchor-lock skip optimization) and the per-line check (for intra-file cap enforcement):

```rust
// File-level: skip anchor work entirely for post-cap files.
if all_results.len() >= max_results {
    truncated += matched_line_nums.len();
    continue;
}

// ... anchor lock + get_anchors (only reached when under cap) ...

// Per-line: enforce cap within a file that straddles the boundary.
for line_num in &matched_line_nums {
    if all_results.len() >= max_results {
        truncated += 1;
    } else {
        let idx = line_num.saturating_sub(1) as usize;
        if let Some((anchor, line)) = anchors.get(idx) {
            all_results.push(format!("{}:{anchor}│{line}\n", relative_path.display()));
        }
    }
}
```

This preserves the anchor-lock optimization for post-cap files while correctly capping output within any single file.

---

A test covering this case would also be useful — `max_results: 2` against a file with 5 matching lines, asserting `all_results` has 2 lines and the truncation message mentions 3 omitted.

No other issues found in this revision.

decko:
Round 3 finding fixed in 8a2f238:

**Regression:** Single file with 5 matches × `max_results=2` → all 5 leaked past the cap because the file-level guard only catches post-cap files, not mid-cap.

**Fix:** Restored per-line guard inside the inner loop. Both guards now coexist:
- **File-level** — skips expensive anchor-lock work for files beyond the cap
- **Per-line** — enforces cap within a file that straddles the boundary

**New test:** `respects_max_results_within_single_file` — 5-match file, `max_results=2`, asserts exactly 2 results + truncation mentions 3 omitted.

`cargo build` ✅ `cargo test` (159) ✅ `cargo clippy` ✅ `cargo fmt` ✅

decko:
## Round 4 Review — Clean Bill

All round-3 findings are correctly addressed in 8a2f238. No new issues found.

**Dual-guard verified correct:**
- File-level check (before anchor lock): post-cap files skip `anchor_state` entirely, count all their matched lines as truncated
- Per-line check (inside inner loop): files straddling the boundary are capped mid-file, excess lines counted individually
- New test `respects_max_results_within_single_file` correctly exercises the boundary case and will catch any future regression

**One minor note for the record (non-blocking):**

The grep scan (`searcher.search_path`) still runs on every file before the file-level cap check. So post-cap files are still read and regex-matched — only the anchor work is skipped. This is the correct tradeoff: skipping the scan would mean losing the accurate truncation count. Just worth a comment in the code if this becomes a perf concern.

**JSON Schema** could add `"minimum": 1` to the `max_results` property to match the `.max(1)` floor enforced in code, but the behavior is already correct without it.

---

Implementation is correct, idiomatic, well-tested (10 tests, 159 total), and consistent with `ReadFileTool` conventions. **Approving.**

PR #77, opened 2026-05-14
feat(tools): add execute_command tool (sandboxed) (#26)

### LINKED ISSUE #26: E4: execute_command tool (sandboxed)
## Context
`src/tools/exec.rs`: Execute shell commands with sandboxing constraints.

**Design ref:** `docs/designs/2026-04-25-carv-design.md`, execute_command Sandboxing section.

## Scope
- 30s timeout (kills child process on expiry)
- Working directory pinned to project root
- Output capped at 32KB (stdout+stderr)
- No shell interpretation: `tokio::process::Command` with arg vector
- Returns exit code, stdout, stderr, truncated flag

**Type:** 🟢 Formulaic (~130 lines)

**Test:** Unit: echo hello → captured; sleep 31 → timeout; output > 32KB → truncated

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
- Implements `ExecuteCommandTool` — sandboxed command execution
- 30s timeout, workspace-root pinned cwd, 32KB output cap, no shell interpretation
- Uses `tokio::process::Command` with arg vector (no `sh -c`)
- 10 tests covering echo, sleep, args, missing/empty command, truncation, metadata

## Changes
| File | Lines | Description |
|------|-------|-------------|
| `src/tools/exec.rs` | +401 | New `ExecuteCommandTool` implementation + test suite |
| `src/tools/mod.rs` | +2 | Module declaration + tool registration |

## Sandbox invariants
- ✅ 30s timeout — `tokio::time::timeout(Duration::from_secs(30), ...)`
- ✅ Pinned cwd — `cmd.current_dir(&ctx.workspace_root)`
- ✅ 32KB output cap — char-boundary-aware truncation
- ✅ No shell — `Command::new(command).args(&args)`, `stdin(Stdio::null())`
- ✅ Kill + reap on timeout — `start_kill()` + `wait().await`
- ✅ Concurrent pipe draining — separate spawned tasks prevent deadlocks
- ✅ No panics — all `.expect()` calls replaced with `Result`/`unwrap_or_default`

## Dependencies
No new dependencies — `tokio` already had `process` feature.

## Checks
- [x] `cargo build` ✅
- [x] `cargo test` (183 tests) ✅
- [x] DoD review — 3 items fixed (panics → Result, metadata test)

Closes #26

### DIFF
--- a/docs/designs/2026-04-25-carv-design.md
+++ b/docs/designs/2026-04-25-carv-design.md
@@ -39,7 +39,7 @@ src/
 │   ├── traits.rs         # Tool trait definition
 │   ├── fs.rs             # read_file, write_file, list_files (.gitignore-aware)
 │   ├── edit.rs           # edit_file (hash-anchored, multi-file batched edits)
-│   ├── exec.rs           # execute_command (sandboxed: timeout, cwd, output cap)
+│   ├── exec.rs           # execute_command (resource-limited: timeout, cwd, output cap)
 │   ├── treesitter.rs     # get_skeleton, get_function, replace_symbol
 │   ├── lsp.rs            # lsp_rename, lsp_references, lsp_definition, lsp_diagnostics
 │   └── search.rs         # search_files (.gitignore-aware, ripgrep-based)
@@ -204,7 +204,7 @@ All tools auto-approved. `--disallowed-tools` removes tools from the registry be
 | `edit_file` | no | Hash-anchored edits: replace/insert_before/insert_after. Supports multi-file batching. |
 | `list_files` | yes | List directory contents (.gitignore-aware) |
 | `search_files` | yes | Ripgrep content search, .gitignore-aware (returns hash-anchored lines) |
-| `execute_command` | no | Run shell command (sandboxed: 30s timeout, cwd pinned, output capped) |
+| `execute_command` | no | Run shell command (resource-limited: 30s timeout, cwd pinned, output capped) |
 | `get_skeleton` | yes | AST structural outline (hash-anchored) |
 | `get_function` | yes | Extract function body by name (hash-anchored) |
 | `replace_symbol` | no | Replace function/class by AST node. Supports multi-file batching. |
--- a/src/tools/exec.rs
+++ b/src/tools/exec.rs
@@ -93,6 +93,11 @@ impl Tool for ExecuteCommandTool {
                 .stdout(std::process::Stdio::piped())
                 .stderr(std::process::Stdio::piped())
                 .stdin(std::process::Stdio::null())
+                // Clear all environment variables to prevent secrets (API keys,
+                // tokens, etc.) from leaking to child processes. Pass only PATH
+                // and HOME; locale vars (LANG, LC_ALL) are intentionally dropped
+                // — some tools may render non-ASCII differently, but the LLM can
+                // observe and adapt.
                 .env_clear()
                 .env("PATH", std::env::var("PATH").unwrap_or_default())
                 .env("HOME", std::env::var("HOME").unwrap_or_default());
--- a/src/tools/exec.rs
+++ b/src/tools/exec.rs
@@ -1,0 +1,401 @@
+//! `execute_command` tool — sandboxed command execution.
+//!
+//! Runs shell commands with a 30-second timeout, working directory pinned to
+//! the workspace root, combined stdout+stderr capped at 32 KB, and no shell
+//! interpretation. Returns exit code, stdout, stderr, and a truncation flag.
+
+use std::time::Duration;
+
+use serde_json::Value;
+use tokio::io::AsyncReadExt;
+use tokio::process::Command;
+
+use crate::tools::traits::{Tool, ToolContext, ToolFuture, ToolResult};
+
+/// Maximum combined stdout+stderr output in bytes before truncation.
+const OUTPUT_CAP: usize = 32 * 1024; // 32 KB
+
+/// Tool that executes a command with sandboxing constraints.
+///
+/// The LLM can use this to run build commands, tests, formatters, linters, etc.
+/// All commands are pinned to the workspace root and cannot escape via `cd` or
+/// shell metacharacters (no shell is involved — arguments are passed directly).
+pub struct ExecuteCommandTool;
+
+impl Tool for ExecuteCommandTool {
+    fn name(&self) -> &str {
+        "execute_command"
+    }
+
+    fn description(&self) -> &str {
+        "Execute a command in the project workspace with sandboxing. \
+         The command runs with a 30-second timeout, working directory pinned \
+         to the workspace root, combined stdout+stderr output capped at 32 KB, \
+         and no shell interpretation (use command + args array — not a shell \
+         string). Returns exit code, stdout, stderr, and a truncation notice \
+         if the output exceeds the cap."
+    }
+
+    fn parameters_schema(&self) -> Value {
+        serde_json::json!({
+            "type": "object",
+            "properties": {
+                "command": {
+                    "type": "string",
+                    "description": "Command to execute (e.g., 'cargo', 'python', 'ls')"
+                },
+                "args": {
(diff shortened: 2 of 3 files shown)


decko:
## Code Review — `execute_command` tool (Rust + SE perspective)

### Overview

Solid implementation of a command-execution tool. The async pipeline design is sound (concurrent pipe draining prevents the classic deadlock where reading stdout blocks while the process fills its stderr buffer), and the no-panic policy is well-executed. A few issues below range from correctness risks to polish.

---

### Issues

#### 1. Abandoned tasks on timeout / wait-error paths — medium

After a timeout or `wait()` I/O error, `stdout_task` and `stderr_task` are never awaited:

```rust
Err(_elapsed) => {
    let _ = child.start_kill();
    let _ = child.wait().await;
    Ok(ToolResult::error("command timed out after 30s and was killed"))
    // stdout_task and stderr_task are dropped here — tasks leak until tokio cleans up
}
```

`start_kill()` sends SIGKILL to the process, which closes its end of the pipes, so the tasks will eventually finish — but they'll hold their `Vec<u8>` allocations until the runtime drops them. In a busy agent loop, many concurrent timeouts could accumulate memory. Consider aborting them explicitly:

```rust
stdout_task.abort();
stderr_task.abort();
let _ = child.start_kill();
let _ = child.wait().await;
```

#### 2. `unwrap_or_default` silently swallows task panics — minor

```rust
let stdout_bytes = stdout_task.await.unwrap_or_default();
```

If the task panics (extremely unlikely here since `read_to_end` doesn't panic), this returns empty bytes without any indication something went wrong. Prefer propagating with `?` or at minimum logging:

```rust
let stdout_bytes = stdout_task.await.map_err(|e| anyhow::anyhow!("stdout task panicked: {e}"))??;
```

#### 3. Environment variables leak to child processes — security concern

The spawned process inherits all environment variables from the parent, including anything the agent was started with: `ANTHROPIC_API_KEY`, `AWS_SECRET_ACCESS_KEY`, database URLs, tokens, etc. A misbehaving command (or LLM-injected argument) could read `env` and exfiltrate secrets.

Suggestion — clear the environment and pass only a minimal safe set:

```rust
cmd.env_clear()
   .env("PATH", std::env::var("PATH").unwrap_or_default())
   .env("HOME", std::env::var("HOME").unwrap_or_default());
```

#### 4. "Sandboxed" is a misleading term — documentation concern

The PR title, description, module docstring, and `description()` all use "sandboxed." In security terminology, sandboxing implies isolation from the host: restricted filesystem access, no network, no arbitrary process spawning. What this tool actually provides is:

- 30s timeout
- CWD pinned to workspace root (but `cd /` via a sub-shell could escape — except there's no shell, so this holds)
- No shell metacharacter injection
- 32 KB output cap

The process can still: read/write any file the agent user can access, make arbitrary network connections, spawn sub-processes, and consume CPU/memory. The description should be honest about this. Consider replacing "sandboxed" with "constrained" or "resource-limited."

#### 5. `output.trim_end().to_string()` allocates unnecessarily — minor

```rust
Ok(ToolResult::ok(output.trim_end().to_string()))
```

`trim_end()` borrows and returns a `&str`; `.to_string()` clones the bytes. Since `output` is already owned, truncate in place:

```rust
let trimmed_len = output.trim_end().len();
output.truncate(trimmed_len);
Ok(ToolResult::ok(output))
```

#### 6. Truncation cap applies to formatted output, not raw data — documentation gap

The 32 KB `OUTPUT_CAP` is measured against the fully formatted string (`"Exit code: N\n--- stdout ---\n...\n--- stderr ---\n..."`), not against raw stdout+stderr bytes. So the effective cap on actual program output is slightly less than 32 KB (minus ~40 bytes of headers). Not a bug, but the constant's doc comment claims "Maximum combined stdout+stderr output in bytes" which is inaccurate. Worth clarifying.

#### 7. `floor_char_boundary` is available since Rust 1.65 — simplification

The truncation boundary search:

```rust
let cap = (0..=OUTPUT_CAP)
    .rev()
    .find(|&i| output.is_char_boundary(i))
    .unwrap_or(0);
```

This works correctly (the range is lazy; UTF-8 chars are ≤4 bytes so at most 4 iterations), but `str::floor_char_boundary` (stable since Rust 1.65) makes the intent clearer:

```rust
let cap = output.floor_char_boundary(OUTPUT_CAP);
output.truncate(cap);
```

#### 8. Platform-specific tests not gated — portability

The tests use `dd`, `sleep`, and `echo` (with Unix semantics). On Windows these don't exist. While the tool itself uses `tokio::process::Command` portably, the tests would fail to compile or run on Windows. Gate them with `#[cfg(unix)]`:

```rust
#[cfg(unix)]
#[tokio::test]
async fn output_truncation() { ... }
```

---

### What's working well

- **Concurrent pipe draining** via separate tokio tasks — correct solution to the stdout/stderr deadlock problem. Well done.
- **`start_kill()` + `wait().await` on timeout** — properly reaps the zombie instead of just forgetting the process.
- **No shell interpretation** — `Command::new(cmd).args(&args)` is the right approach; no `sh -c` injection surface.
- **`stdin(Stdio::null())`** — prevents interactive prompts from hanging.
- **UTF-8 boundary-aware truncation** — correct; avoids splitting multi-byte characters.
- **No panics** — all `.expect()` removed; errors flow through `ToolResult::error`.
- **Test for `command` being a non-string JSON value** — good edge case coverage.

decko:
7 of 8 findings addressed in a55ad50:

| # | Status | Fix |
|---|--------|-----|
| 1 | ✅ | stdout_task.abort() + stderr_task.abort() on both error/timeout paths |
| 2 | ✅ | tracing::warn!() via inspect_err before unwrap_or_default |
| 3 | ✅ | env_clear() + minimal PATH/HOME — no secret leakage |
| 4 | ✅ | 'sandboxed' → 'resource-limited' in exec.rs and design doc |
| 5 | ✅ | In-place truncate instead of trim_end().to_string() |
| 6 | ✅ | OUTPUT_CAP doc clarified: formatted output, not raw bytes |
| 7 | ⏸️ | floor_char_boundary not stable in Rust 1.86 — deferred |
| 8 | ✅ | #[cfg(unix)] on echo, sleep, args, truncation tests |

cargo build ✅ cargo test (183) ✅ cargo clippy ✅ cargo fmt ✅

decko:
Re: finding #7 — `str::floor_char_boundary` is still feature-gated on Rust 1.86 (tracking issue #93743). The manual boundary search is correct and worst-case 4 iterations for a 4-byte UTF-8 character. Will switch when it lands in stable.

decko:
## Follow-up review — post-fix verification

Confirmed: all 7 applied fixes look correct in `a55ad50`. The code is clean.

---

### Correction on finding #7

My claim that `str::floor_char_boundary` is "stable since Rust 1.65" was **wrong**. Just confirmed: on Rust 1.86 stable it errors with `E0658: use of unstable library feature 'round_char_boundary'` (tracking issue #93743, still open). The manual boundary search is the correct approach and the only one available on stable. Apologies for the noise — your deferral was right.

---

### One new observation — `env_clear` drops locale variables

With the env fix applied:

```rust
cmd.env_clear()
   .env("PATH", std::env::var("PATH").unwrap_or_default())
   .env("HOME", std::env::var("HOME").unwrap_or_default());
```

`LANG` and `LC_ALL` are stripped. Some tools change behavior without them:
- `sort` may not handle UTF-8 ranges correctly
- Error messages from certain programs fall back to C/POSIX locale
- `git` may render non-ASCII author names as escape sequences

Not a blocker — the LLM will observe failure and adapt — but worth documenting in the `description()` string or a code comment so it's not a surprise when `cargo test` output looks different inside the tool vs. the terminal.

---

### Verdict

All substantive issues from the first review are resolved. The two remaining items (#7 nit retracted, locale note above) are documentation-level at most. **LGTM.**

PR #83, opened 2026-05-19
F2: edit_file — insert_before + insert_after operations (#28)

### LINKED ISSUE #28: F2: edit_file — insert_before + insert_after operations

decko:
## Summary
Add `insert_before` and `insert_after` operations to the edit_file tool.

## Changes
- Add `operation` parameter (enum: `replace`/default, `insert_before`, `insert_after`)
- Relax `end_anchor` from required to optional (only required for `replace`)
- New helpers: `insert_lines_before()` and `insert_lines_after()`
- `insert_lines_after` delegates to `insert_lines_before(at + 1)`
- Anchor cache invalidation on insert (same as replace)
- 7 new tests: 5 unit + 3 integration (incl. cache invalidation)
- Backward compatible: omit `operation` → behaves as `replace`

## Verification
- ✅ `cargo build`
- ✅ `cargo test` (208 passed)
- ✅ `cargo clippy -- -D warnings`
- ✅ `cargo fmt -- --check`
- ✅ DoD reviewed by rust-reviewer

Closes #28

### DIFF
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -499,6 +499,13 @@ mod tests {
         assert_eq!(replace_line_range(content, 1, 1, "X"), "a\nX\nc\n");
     }
 
+    #[test]
+    fn replace_double_newline_appends_blank_line() {
+        // "\n\n" in new_text keeps content + appends a trailing blank line.
+        let content = "a\nb\nc\n";
+        assert_eq!(replace_line_range(content, 1, 1, "X\n\n"), "a\nX\n\nc\n");
+    }
+
     // -----------------------------------------------------------------------
     // EditFileTool integration tests
     // -----------------------------------------------------------------------
@@ -794,6 +801,13 @@ mod tests {
         // Both inputs produce the same concrete output — trailing \n is stripped.
         assert_eq!(insert_lines_before(content, 1, "foo\n"), "a\nfoo\nb\n");
         assert_eq!(insert_lines_before(content, 1, "foo"), "a\nfoo\nb\n");
+    }
+
+    #[test]
+    fn insert_double_newline_appends_blank_line() {
+        // "\n\n" keeps one line of content + appends a trailing blank line.
+        let content = "a\nb\n";
+        assert_eq!(insert_lines_before(content, 1, "foo\n\n"), "a\nfoo\n\nb\n");
     }
 
     // -----------------------------------------------------------------------
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -8,6 +8,33 @@ use crate::tools::traits::{Tool, ToolContext, ToolFuture, ToolResult};
 use serde_json::Value;
 use std::path::{Path, PathBuf};
 
+// ---------------------------------------------------------------------------
+// Operation enum
+// ---------------------------------------------------------------------------
+
+/// Operation dispatched by [`EditFileTool`].
+///
+/// Parsed once from the `operation` field in the tool input; all downstream
+/// matches are exhaustive, so the compiler catches a missing arm whenever a
+/// new variant is added.
+#[derive(Copy, Clone, Debug, PartialEq)]
+enum EditOp {
+    Replace,
+    InsertBefore,
+    InsertAfter,
+}
+
+impl EditOp {
+    fn from_str(s: &str) -> Option<Self> {
+        match s {
+            "replace" => Some(Self::Replace),
+            "insert_before" => Some(Self::InsertBefore),
+            "insert_after" => Some(Self::InsertAfter),
+            _ => None,
+        }
+    }
+}
+
 /// Tool for anchor-based file editing.
 ///
 /// Supports three operations: `replace` (default), `insert_before`, and
@@ -69,15 +96,25 @@ impl Tool for EditFileTool {
                 Some(p) => p,
                 None => return Ok(ToolResult::error("missing required 'path' parameter")),
             };
-            let operation = input
+            let operation_str = input
                 .get("operation")
                 .and_then(Value::as_str)
                 .unwrap_or("replace");
+            let op = match EditOp::from_str(operation_str) {
+                Some(op) => op,
+                None => {
+                    return Ok(ToolResult::error(format!(
+                        "edit_file failed: unknown operation '{}'. \
+                         Valid: replace, insert_before, insert_after",
+                        operation_str
+                    )))
+                }
+            };
             let anchor = match input.get("anchor").and_then(Value::as_str) {
                 Some(a) => a,
                 None => return Ok(ToolResult::error("missing required 'anchor' parameter")),
             };
-            let end_anchor = if operation == "replace" {
+            let end_anchor = if op == EditOp::Replace {
                 match input.get("end_anchor").and_then(Value::as_str) {
                     Some(a) => a,
                     None => {
@@ -137,7 +174,7 @@ impl Tool for EditFileTool {
                     }
                 };
 
-                if operation == "replace" {
+                if op == EditOp::Replace {
                     let end = match anchors.iter().position(|(a, _)| a == end_anchor) {
                         Some(idx) => idx,
                         None => {
@@ -179,7 +216,7 @@ impl Tool for EditFileTool {
 
             // Guard: ensure referenced lines exist in the current file content.
             let line_count = content.lines().count();
-            if operation == "replace" {
+            if op == EditOp::Replace {
                 if end_line >= line_count {
                     return Ok(ToolResult::error(format!(
                         "edit_file failed: end_anchor '{}' refers to line {} but \
@@ -200,21 +237,10 @@ impl Tool for EditFileTool {
             }
 
             // Apply the edit based on operation.
-            let (new_content, lines_changed) = match operation {
-                "insert_before" => (
-                    insert_lines_before(&content, start_line, text),
-                    text.lines().count(),
-                ),
-                "insert_after" => (
-                    insert_lines_after(&content, start_line, text),
-                    text.lines().count(),
-                ),
-                // "replace" is the default; the schema enum prevents unknown
-                // values in practice, but falling back to replace is safe.
-                _ => (
-                    replace_line_range(&content, start_line, end_line, text),
-                    end_line - start_line + 1,
-                ),
+            let new_content = match op {
+                EditOp::Replace => replace_line_range(&content, start_line, end_line, text),
+                EditOp::InsertBefore => insert_lines_before(&content, start_line, text),
+                EditOp::InsertAfter => insert_lines_after(&content, start_line, text),
             };
 
             // Write back.
@@ -230,32 +256,30 @@ impl Tool for EditFileTool {
                         anchor_state.notify_edit(&resolved);
                     }
 
-                    let msg = if operation == "replace" {
-                        format!(
+                    let msg = match op {
+                        EditOp::Replace => format!(
                             "Replaced {} line(s) ({}→{} bytes) in {}",
-                            lines_changed,
+                            end_line - start_line + 1,
                             old_bytes,
                             new_bytes,
                             resolved.display()
-                        )
-                    } else if operation == "insert_before" {
-                        format!(
+                        ),
+                        EditOp::InsertBefore => format!(
                             "Inserted {} line(s) before anchor '{}' ({}→{} bytes) in {}",
-                            lines_changed,
+                            text.lines().count(),
                             anchor,
                             old_bytes,
                             new_bytes,
                             resolved.display()
-                        )
-                    } else {
-                        format!(
+                        ),
+                        EditOp::InsertAfter => format!(
                             "Inserted {} line(s) after anchor '{}' ({}→{} bytes) in {}",
-                            lines_changed,
+                            text.lines().count(),
                             anchor,
                             old_bytes,
                             new_bytes,
                             resolved.display()
-                        )
+                        ),
                     };
                     Ok(ToolResult::ok(msg))
                 }
@@ -279,6 +303,12 @@ impl Tool for EditFileTool {
 /// no-op (returns `content` unchanged).
 fn insert_lines_before(content: &str, at: usize, new_text: &str) -> String {
     let lines: Vec<&str> = content.lines().collect();
+    debug_assert!(
+        at <= lines.len(),
+        "insert_lines_before: at {} out of bounds ({} lines)",
+        at,
+        lines.len()
+    );
     let new_lines: Vec<&str> = new_text.lines().collect();
     if new_lines.is_empty() {
         return content.to_string();
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -79,7 +79,7 @@ impl Tool for EditFileTool {
                 },
                 "text": {
                     "type": "string",
-                    "description": "New text to insert or replace with. Include newline characters between lines as needed."
+                    "description": "New text to insert or replace with. Include newline characters between lines as needed. For insert operations, a trailing newline is treated as a line terminator and stripped; to also insert a trailing blank line, end with '\\n\\n'."
                 }
             },
             "required": ["path", "anchor", "text"]
@@ -278,22 +278,22 @@ impl Tool for EditFileTool {
                             new_bytes,
                             resolved.display()
                         ),
-                        EditOp::InsertBefore => format!(
-                            "Inserted {} line(s) before anchor '{}' ({}→{} bytes) in {}",
-                            text.lines().count(),
-                            anchor,
-                            old_bytes,
-                            new_bytes,
-                            resolved.display()
-                        ),
-                        EditOp::InsertAfter => format!(
-                            "Inserted {} line(s) after anchor '{}' ({}→{} bytes) in {}",
-                            text.lines().count(),
-                            anchor,
-                            old_bytes,
-                            new_bytes,
-                            resolved.display()
-                        ),
+                        EditOp::InsertBefore | EditOp::InsertAfter => {
+                            let dir = if op == EditOp::InsertBefore {
+                                "before"
+                            } else {
+                                "after"
+                            };
+                            format!(
+                                "Inserted {} line(s) {} anchor '{}' ({}→{} bytes) in {}",
+                                text.lines().count(),
+                                dir,
+                                anchor,
+                                old_bytes,
+                                new_bytes,
+                                resolved.display()
+                            )
+                        }
                     };
                     Ok(ToolResult::ok(msg))
                 }
@@ -452,6 +452,13 @@ mod tests {
         let content = "keep\nremove\nkeep\n";
         let result = replace_line_range(content, 1, 1, "");
         assert_eq!(result, "keep\n\nkeep\n");
+    }
+
+    #[test]
+    fn replace_single_line_with_empty_collapses_file() {
+        // Known edge case: sole line + empty replacement → empty file.
+        // (Multi-line files preserve the blank line; single-line collapses.)
+        assert_eq!(replace_line_range("only\n", 0, 0, ""), "");
     }
 
     #[test]
@@ -770,10 +777,9 @@ mod tests {
         // "foo\n" and "foo" produce the same insert. To append a blank
         // line, end with "\n\n".
         let content = "a\nb\n";
-        assert_eq!(
-            insert_lines_before(content, 1, "foo\n"),
-            insert_lines_before(content, 1, "foo")
-        );
+        // Both inputs produce the same concrete output — trailing \n is stripped.
+        assert_eq!(insert_lines_before(content, 1, "foo\n"), "a\nfoo\nb\n");
+        assert_eq!(insert_lines_before(content, 1, "foo"), "a\nfoo\nb\n");
     }
 
     // -----------------------------------------------------------------------
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -130,7 +130,6 @@ impl Tool for EditFileTool {
                 Some(t) => t,
                 None => return Ok(ToolResult::error("missing required 'text' parameter")),
             };
-            let text_line_count = text.lines().count();
 
             // Reject empty text for insert operations — "insert nothing" is
             // almost certainly a caller mistake. (Replace treats empty text as
@@ -281,7 +280,7 @@ impl Tool for EditFileTool {
                         ),
                         EditOp::InsertBefore => format!(
                             "Inserted {} line(s) before anchor '{}' ({}→{} bytes) in {}",
-                            text_line_count,
+                            text.lines().count(),
                             anchor,
                             old_bytes,
                             new_bytes,
@@ -289,7 +288,7 @@ impl Tool for EditFileTool {
                         ),
                         EditOp::InsertAfter => format!(
                             "Inserted {} line(s) after anchor '{}' ({}→{} bytes) in {}",
-                            text_line_count,
+                            text.lines().count(),
                             anchor,
                             old_bytes,
                             new_bytes,
@@ -318,7 +317,10 @@ impl Tool for EditFileTool {
 /// no-op (returns `content` unchanged).
 ///
 /// Note: `str::lines()` normalizes `\r\n` → `\n`, so Windows-style line
-/// endings are converted to LF on write.
+/// endings are converted to LF on write. A trailing `\n` in `new_text` is
+/// treated as a line terminator and stripped (consistent with `lines()`
+/// semantics). To append a blank line after the last inserted line, end
+/// `new_text` with `"\n\n"`.
 fn insert_lines_before(content: &str, at: usize, new_text: &str) -> String {
     let lines: Vec<&str> = content.lines().collect();
     debug_assert!(
@@ -760,6 +762,18 @@ mod tests {
         let content = "a\r\nb\r\n";
         let result = insert_lines_before(content, 0, "X");
         assert_eq!(result, "X\na\nb\n");
+    }
+
+    #[test]
+    fn insert_trailing_newline_in_text_stripped() {
+        // str::lines() treats a trailing \n as a terminator, not content.
+        // "foo\n" and "foo" produce the same insert. To append a blank
+        // line, end with "\n\n".
+        let content = "a\nb\n";
+        assert_eq!(
+            insert_lines_before(content, 1, "foo\n"),
+            insert_lines_before(content, 1, "foo")
+        );
     }
 
     // -----------------------------------------------------------------------
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -10,8 +10,9 @@ use std::path::{Path, PathBuf};
 
 /// Tool for anchor-based file editing.
 ///
-/// Replaces an inclusive range of lines — identified by anchor words from
-/// [`read_file`](crate::tools::fs::ReadFileTool) — with new text.
+/// Supports three operations: `replace` (default), `insert_before`, and
+/// `insert_after`. Uses stable word anchors from
+/// [`read_file`](crate::tools::fs::ReadFileTool) to target edit locations.
 pub struct EditFileTool;
 
 impl Tool for EditFileTool {
@@ -208,6 +209,8 @@ impl Tool for EditFileTool {
                     insert_lines_after(&content, start_line, text),
                     text.lines().count(),
                 ),
+                // "replace" is the default; the schema enum prevents unknown
+                // values in practice, but falling back to replace is safe.
                 _ => (
                     replace_line_range(&content, start_line, end_line, text),
                     end_line - start_line + 1,
@@ -297,7 +300,9 @@ fn insert_lines_before(content: &str, at: usize, new_text: &str) -> String {
 /// Same semantics as [`insert_lines_before`], but inserts after `at` instead
 /// of before. An empty `new_text` is a no-op.
 fn insert_lines_after(content: &str, at: usize, new_text: &str) -> String {
-    insert_lines_before(content, at + 1, new_text)
+    // saturating_add guards against usize::MAX overflow (not reachable
+    // in practice — the caller guards at < line_count).
+    insert_lines_before(content, at.saturating_add(1), new_text)
 }
 
 /// Replace lines from `start_line` to `end_line` (inclusive, 0-indexed) in
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -130,12 +130,12 @@ impl Tool for EditFileTool {
                 Some(t) => t,
                 None => return Ok(ToolResult::error("missing required 'text' parameter")),
             };
+            let text_line_count = text.lines().count();
 
             // Reject empty text for insert operations — "insert nothing" is
             // almost certainly a caller mistake. (Replace treats empty text as
             // "blank the line", which is a meaningful operation.)
-            if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text.lines().count() == 0
-            {
+            if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text.is_empty() {
                 return Ok(ToolResult::error(
                     "'text' must be non-empty for insert operations",
                 ));
@@ -281,7 +281,7 @@ impl Tool for EditFileTool {
                         ),
                         EditOp::InsertBefore => format!(
                             "Inserted {} line(s) before anchor '{}' ({}→{} bytes) in {}",
-                            text.lines().count(),
+                            text_line_count,
                             anchor,
                             old_bytes,
                             new_bytes,
@@ -289,7 +289,7 @@ impl Tool for EditFileTool {
                         ),
                         EditOp::InsertAfter => format!(
                             "Inserted {} line(s) after anchor '{}' ({}→{} bytes) in {}",
-                            text.lines().count(),
+                            text_line_count,
                             anchor,
                             old_bytes,
                             new_bytes,
@@ -671,6 +671,43 @@ mod tests {
         assert!(result.content.contains("File may have changed"));
     }
 
+    #[tokio::test]
+    async fn stale_anchor_before_insert() {
+        let dir = temp_dir("edit_stale_insert_anchor");
+        let file = write_temp_file(&dir, "stale.rs", "line1\nline2\nline3\n");
+        let ctx = test_context(dir.clone());
+
+        // Prime the anchor cache.
+        let anchors = {
+            let mut state = ctx.anchor_state.lock().unwrap();
+            state.get_anchors(&file).unwrap()
+        };
+        let anchor_line3 = &anchors[2].0; // third line's anchor
+
+        // Truncate the file externally to one line. The anchor cache
+        // still reports 3 lines; insert_before should catch the mismatch
+        // via its start_line >= line_count guard.
+        std::fs::write(&file, "only one line\n").unwrap();
+
+        let tool = EditFileTool;
+        let result = tool
+            .execute(
+                json!({
+                    "path": file.to_str().unwrap(),
+                    "operation": "insert_before",
+                    "anchor": anchor_line3,
+                    "text": "inserted"
+                }),
+                &ctx,
+            )
+            .await
+            .unwrap();
+
+        assert!(result.is_error);
+        assert!(result.content.contains("file has"));
+        assert!(result.content.contains("File may have changed"));
+    }
+
     // -----------------------------------------------------------------------
     // insert_lines_before / insert_lines_after unit tests
     // -----------------------------------------------------------------------
@@ -874,19 +911,42 @@ mod tests {
         let file = write_temp_file(&dir, "target.rs", "a\nb\n");
         let ctx = test_context(dir.clone());
 
-        let anchors = {
-            let mut state = ctx.anchor_state.lock().unwrap();
-            state.get_anchors(&file).unwrap()
-        };
-        let (anchor, _) = &anchors[0];
-
+        // The empty-text guard fires before path resolution or anchor
+        // lookup — any anchor string works. We pass a dummy to keep the
+        // test focused on parameter validation.
         let tool = EditFileTool;
         let result = tool
             .execute(
                 json!({
                     "path": file.to_str().unwrap(),
                     "operation": "insert_before",
-                    "anchor": anchor,
+                    "anchor": "any-word",
+                    "text": ""
+                }),
+                &ctx,
+            )
+            .await
+            .unwrap();
+
+        assert!(result.is_error);
+        assert!(result
+            .content
+            .contains("'text' must be non-empty for insert operations"));
+    }
+
+    #[tokio::test]
+    async fn insert_after_empty_text_rejected() {
+        let dir = temp_dir("edit_insert_after_empty");
+        let file = write_temp_file(&dir, "target.rs", "a\nb\n");
+        let ctx = test_context(dir.clone());
+
+        let tool = EditFileTool;
+        let result = tool
+            .execute(
+                json!({
+                    "path": file.to_str().unwrap(),
+                    "operation": "insert_after",
+                    "anchor": "any-word",
                     "text": ""
                 }),
                 &ctx,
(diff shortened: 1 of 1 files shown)


decko (commented):
## Rust Specialist Review — PR #83: `insert_before` / `insert_after`

The implementation is functionally correct and the tests are solid. One structural issue deserves attention before merge; the rest are minor.

---

### Critical: Operation dispatched via repeated string comparisons

The `operation: &str` value is compared against literal strings **six or more times** across the function body (parameter extraction, end-anchor gating, line-count guard, dispatch match, success message, and implicitly in `lines_changed` semantics). This pattern has two concrete risks:

1. **Silent fall-through**: The `_ =>` arm of the dispatch `match` silently treats any unknown operation as `replace`. The comment acknowledges the schema enum prevents this, but that assumption breaks the moment the tool is called programmatically or a new variant is added mid-PR. An unknown `"upsert"` would silently overwrite.

2. **Scaling cost**: Adding a fourth operation requires touching every comparison site. It's easy to miss one.

The idiomatic fix is to parse the string **once** into a local enum at the call site, then propagate the enum:

```rust
#[derive(Copy, Clone)]
enum EditOp { Replace, InsertBefore, InsertAfter }

impl EditOp {
    fn from_str(s: &str) -> Option<Self> {
        match s {
            "replace"       => Some(Self::Replace),
            "insert_before" => Some(Self::InsertBefore),
            "insert_after"  => Some(Self::InsertAfter),
            _               => None,
        }
    }
}
```

Then:
```rust
let op = match input.get("operation").and_then(Value::as_str).unwrap_or("replace") {
    s => EditOp::from_str(s)
        .ok_or_else(|| /* return ToolResult::error */)?,
};
```

All downstream matches become exhaustive and the compiler catches any missed arm when a new variant is added.

---

### Minor: `lines_changed` counts different things for replace vs. insert

For `replace`, `lines_changed = end_line - start_line + 1` counts **source lines removed**. For insert ops, it's `text.lines().count()` — **new lines added**. Both values happen to appear in correct messages, but the variable name implies a single meaning. Consider `lines_replaced` / `lines_inserted` or just inline the expressions directly in the message format strings to make intent obvious.

---

### Minor: `insert_lines_before` will panic if `at > lines.len()`

The caller's `start_line < line_count` guard ensures safety for all current call paths. But the helper itself does an unchecked `&lines[..at]` slice without asserting `at <= lines.len()`. Because this function is `pub(crate)` and directly exercised in unit tests, it's one off-by-one away from a panic if a new caller is added without the same guard. A `debug_assert!(at <= lines.len())` or a saturating clamp would document the invariant explicitly.

---

### Positive observations

- `insert_after` delegating to `insert_lines_before(at + 1)` is clean and avoids duplicated logic.
- `saturating_add` on the delegation call is a nice defensive touch.
- Anchor cache invalidation is correctly applied for insert ops (same path as replace).
- The 3 integration tests cover: before-anchor, after-anchor, and cache invalidation — all the cases that matter.
- Backward compatibility via defaulting `operation` to `"replace"` is the right call.
- `insert_empty_text_is_noop` covering both helpers in one test is efficient.

---

### Summary

| Finding | Severity |
|---------|----------|
| Operation dispatched via repeated string literals (silent fall-through risk, no exhaustiveness) | **Should fix** |
| `lines_changed` semantic inconsistency | Nit |
| `insert_lines_before` lacks invariant assertion for `at` bound | Nit |

The enum refactor is the one I'd want addressed before merge; the rest can be follow-ups if preferred.

decko (commented):
All three findings addressed in commit 990bff5:

**1. Enum refactor (the "should fix")** — Added `EditOp` enum with `from_str()` parser. The operation string is parsed once at the call site, then all downstream matches are exhaustive. Unknown operations return an explicit error instead of silently falling through to replace. Adding a fourth variant now requires touching only: the enum definition, `from_str`, `parameters_schema`, and the three `match op { }` arms — the compiler catches any missed arm.

**2. `lines_changed` semantic ambiguity** — Eliminated the variable entirely. The replace count (`end_line - start_line + 1`) and insert count (`text.lines().count()`) are now inlined directly in their respective `EditOp::Replace` / `EditOp::Insert*` format strings. No variable carrying different meanings in different branches.

**3. `insert_lines_before` bounds invariant** — Added `debug_assert!(at <= lines.len(), ...)` at the top of the helper. Documents the contract explicitly and catches violations in debug builds.

Verification: `cargo build` ✅ `cargo test` (208 passed) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 2 (against current branch state)

The previous round's findings have been fully addressed: `EditOp` enum is in, exhaustive `match` is used for dispatch and the success message, unknown operations now return a descriptive error, and `debug_assert!` bounds guards are in both helpers. That's the right shape.

Fresh pass on the current state of `src/tools/edit.rs`:

---

### Finding 1: New error path (`EditOp::from_str` returning `None`) has no test

`EditOp::from_str` now correctly rejects unknown operation strings with a useful error message. But there is no test that exercises this path — i.e., calling `execute` with `"operation": "delete"` or any other non-enum string. The schema enum prevents this from a conforming caller, but the function is correct to defend against it; the defense should be verified.

Suggested test (can be a quick unit-style integration test):

```rust
#[tokio::test]
async fn unknown_operation_rejected() {
    let dir = temp_dir("edit_unknown_op");
    let file = write_temp_file(&dir, "f.rs", "a\n");
    let ctx = test_context(dir.clone());
    let tool = EditFileTool;
    let result = tool.execute(
        json!({"path": file.to_str().unwrap(), "operation": "delete",
               "anchor": "x", "text": "y"}),
        &ctx,
    ).await.unwrap();
    assert!(result.is_error);
    assert!(result.content.contains("unknown operation"));
}
```

---

### Finding 2: Empty `text` + insert operation silently succeeds and reports "Inserted 0 line(s)"

`insert_lines_before` correctly short-circuits and returns `content` unchanged when `new_text` is empty. However, the `execute` path doesn't guard for this before calling the helper, so a caller that passes `"text": ""` with `insert_before` or `insert_after` will:

1. Get `Ok(ToolResult::ok(...))` — no error signal
2. Receive the message `"Inserted 0 line(s) before anchor '...' (N→N bytes) in ..."` — a successful no-op

Compare with `replace`: an empty `text` is a meaningful operation (blank out a line). For insert, "insert nothing" is almost certainly a caller mistake. Consider rejecting it early:

```rust
if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text.trim().is_empty() {
    return Ok(ToolResult::error("'text' must not be empty for insert operations"));
}
```

Or at minimum, a test that documents the current no-op behaviour as intentional, so the next reader knows it wasn't overlooked.

---

### Finding 3: CRLF normalisation not mentioned in `insert_lines_before` docstring

`replace_line_range` has a test (`crlf_normalized_to_lf`) and a comment in that test explicitly documenting the behaviour. `insert_lines_before` uses the same `content.lines()` mechanism and will silently convert `\r\n` → `\n` on Windows-style files. The docstring doesn't mention this. The existing CRLF test coverage doesn't extend to insert operations. Not a correctness bug, but a documentation gap and a missing test for consistency.

---

### Finding 4 (nit): Bounds check for non-replace ops uses `else if` instead of `match`

```rust
if op == EditOp::Replace {
    if end_line >= line_count { ... }
} else if start_line >= line_count { ... }
```

Now that we have an exhaustive enum, a `match` would make the `InsertBefore` / `InsertAfter` arms explicit and catch any future variant at compile time:

```rust
match op {
    EditOp::Replace => {
        if end_line >= line_count { return Err(...); }
    }
    EditOp::InsertBefore | EditOp::InsertAfter => {
        if start_line >= line_count { return Err(...); }
    }
}
```

---

### Positive delta since round 1

- `EditOp` enum fully eliminates the string-comparison proliferation — clean.
- Unknown operations now produce a useful error instead of silently falling through to `replace`.
- `debug_assert!` in both helpers with message format documenting the invariant — exactly right.
- `lines_changed` removed in favour of inline expressions per match arm — semantically unambiguous now.

---

### Summary

| Finding | Severity |
|---------|----------|
| `EditOp::from_str` `None` path untested | Should fix |
| Empty `text` + insert silently no-ops with a misleading success message | Should fix or document |
| CRLF behaviour undocumented / untested for insert helpers | Nit |
| Bounds check could be `match` on `EditOp` | Nit |

decko (commented):
All four findings addressed in commit 2b96a5a:

**Finding 1 — unknown op test** ✅ Added `unknown_operation_rejected` integration test (calls execute with `"operation": "delete"`, asserts error + message contains "unknown operation").

**Finding 2 — empty text + insert** ✅ Now rejected early at the `execute` level with `matches!(op, InsertBefore | InsertAfter) && text.lines().count() == 0`. Returns a clear error: "text must be non-empty for insert operations". Replace still treats empty text as "blank the line" (unchanged). Added `insert_empty_text_rejected` test.

**Finding 3 — CRLF doc/test** ✅ Added note to `insert_lines_before` doc: "Note: `str::lines()` normalizes `\r\n` → `\n`". Added `insert_crlf_normalized_to_lf` test (`"a\r\nb\r\n"` → LF output).

**Finding 4 — bounds check match** ✅ Replaced `if op == Replace { ... } else if start_line >= line_count { ... }` with exhaustive `match op { Replace => ..., InsertBefore | InsertAfter => ... }`.

Test count: 208 → 211 (+3: unknown_operation_rejected, insert_empty_text_rejected, insert_crlf_normalized_to_lf)
Verification: `cargo build` ✅ `cargo test` (211 passed) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 3

All round 2 findings are resolved. Good iteration. The remaining items are small but worth a look before merge.

---

### Finding 1: `insert_empty_text_rejected` test only exercises `insert_before`

The guard at line 137 covers both operations:
```rust
if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text.lines().count() == 0
```
But the test only sends `"operation": "insert_before"`. The `insert_after` branch of the guard is never exercised by the test suite. One additional case covers it:

```rust
#[tokio::test]
async fn insert_after_empty_text_rejected() {
    // ... same setup ...
    let result = tool.execute(
        json!({"path": ..., "operation": "insert_after", "anchor": anchor, "text": ""}),
        &ctx,
    ).await.unwrap();
    assert!(result.is_error);
    assert!(result.content.contains("'text' must be non-empty"));
}
```

---

### Finding 2: No stale-anchor test for insert operations

`stale_anchor_after_external_modification` tests the `Replace` path through the `end_line >= line_count` guard. Insert operations have their own branch in the same match:

```rust
EditOp::InsertBefore | EditOp::InsertAfter => {
    if start_line >= line_count { ... }
}
```

This path is never exercised by a test. If a file is truncated between `read_file` and `insert_before`, the guard should fire — but that's currently unverified. A test mirroring the existing stale-anchor test but using `"operation": "insert_before"` would close the gap.

---

### Finding 3 (nit): `text.lines().count() == 0` should be `text.is_empty()`

Verified: for any non-empty string `s`, `s.lines().count() >= 1` (`"\n"` yields 1, `" "` yields 1, etc.). The two expressions are equivalent. `text.is_empty()` states the intent directly without invoking the line-split machinery, and is the idiomatic Rust form here.

---

### Finding 4 (nit): `text.lines().count()` computed twice for insert ops

The count is computed at line 137 (guard) and again at lines 284 and 291 (success message). Binding it once avoids the redundant traversal:

```rust
let text_line_count = text.lines().count();
if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text_line_count == 0 { ... }
// then use text_line_count in the success message
```

Trivial cost in practice, but cleaner.

---

### Finding 5 (nit): `insert_empty_text_rejected` test primes anchor cache unnecessarily

The test calls `state.get_anchors(&file)` to obtain `anchor` — but the empty-text guard fires at line 137, *before* any path resolution or anchor lookup. The primed anchor word is never consulted. The test could be simplified to pass any arbitrary anchor string, which also makes the test's intent clearer (it's testing parameter validation, not anchor resolution).

---

### Delta since round 2: all resolved ✅

| Round 2 finding | Status |
|----------------|--------|
| `EditOp::from_str` None path untested | ✅ `unknown_operation_rejected` added |
| Empty text + insert silent no-op | ✅ Guard + `insert_empty_text_rejected` added |
| CRLF undocumented for insert helpers | ✅ Docstring note + `insert_crlf_normalized_to_lf` added |
| Bounds check `if/else if` instead of `match` | ✅ Converted to exhaustive `match` |

### Round 3 summary

| Finding | Severity |
|---------|----------|
| `insert_after` empty-text rejection untested | Should fix |
| Stale-anchor path for insert ops untested | Should fix |
| `text.lines().count() == 0` → `text.is_empty()` | Nit |
| `text.lines().count()` computed twice | Nit |
| `insert_empty_text_rejected` primes cache it doesn't use | Nit |

decko (commented):
All five findings addressed in commit c7613f3:

**Finding 1 — insert_after empty-text test** ✅ Added `insert_after_empty_text_rejected`. Now both branches of the `matches!(op, InsertBefore | InsertAfter)` guard are exercised.

**Finding 2 — stale-anchor insert test** ✅ Added `stale_anchor_before_insert`. Mirrors existing replace stale-anchor test. Primes cache for 3-line file, truncates to 1 line externally, then calls insert_before with the stale anchor. Guard fires correctly: `start_line >= line_count`.

**Finding 3 — text.is_empty()** ✅ Replaced `text.lines().count() == 0` with `text.is_empty()` in the guard. Equivalent (verified: non-empty strings have `lines().count() >= 1`), more idiomatic.

**Finding 4 — bind count once** ✅ Added `let text_line_count = text.lines().count();` after text extraction. Used in the success messages instead of calling `text.lines().count()` twice.

**Finding 5 — simplify test** ✅ Removed anchor cache priming from `insert_empty_text_rejected`. Test now passes a dummy anchor string — guard fires before anchor lookup, so priming was misleading.

Test count: 211 → 213 (+2: `insert_after_empty_text_rejected`, `stale_anchor_before_insert`)
Verification: `cargo build` ✅ `cargo test` (213) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 4

All round 3 findings resolved. The code is in genuinely good shape. Two remaining issues found on this pass, both worth addressing before merge.

---

### Finding 1: `text_line_count` computed unconditionally but only used for insert ops

Line 133 binds `text_line_count` before `op` has been used for any branching:

```rust
let text_line_count = text.lines().count();   // line 133 — always runs
// ...
if matches!(op, EditOp::InsertBefore | EditOp::InsertAfter) && text.is_empty() {
    return Ok(...);                            // line 138 — early return for insert+empty
}
```

Two problems:
1. For `EditOp::Replace`, `text_line_count` is computed and then never read. Rust doesn't warn because the variable *is* referenced in the insert arms of the success-message match — but those arms are unreachable for a `Replace` call.
2. For the `insert + empty text` early-return path (line 138), `text_line_count` is computed (it's 0) and then discarded.

Moving the binding to just before the success message, inside the write-success branch, removes both dead computations and makes the read/write locality obvious:

```rust
// in the Ok(()) write branch:
let msg = match op {
    EditOp::Replace => format!("Replaced {} ...", end_line - start_line + 1, ...),
    EditOp::InsertBefore | EditOp::InsertAfter => {
        let line_count = text.lines().count();
        format!("Inserted {} ...", line_count, ...)
    }
};
```

---

### Finding 2: Trailing `\n` in `text` is silently stripped for insert ops — undocumented and untested

`insert_lines_before` collects `new_text.lines()`, and `str::lines()` treats a trailing newline as a line terminator — not as a content character. This means:

| `text` value | Lines inserted |
|---|---|
| `"foo"` | `["foo"]` — 1 line |
| `"foo\n"` | `["foo"]` — 1 line (trailing `\n` stripped) |
| `"foo\n\n"` | `["foo", ""]` — 2 lines (the second `\n` creates a blank) |

So `"foo\n"` and `"foo"` produce identical results. A caller trying to insert a trailing blank line with `"text": "fn foo() {\n}\n"` gets only the function body — not the blank line they expected. They'd need `"text": "fn foo() {\n}\n\n"`.

This is consistent with how `replace_line_range` handles `new_text`, but it's non-obvious and neither the `text` parameter description, the `insert_lines_before` docstring, nor any test documents it.

Suggested docstring addition to `insert_lines_before`:
```
/// A trailing newline in `new_text` is treated as a line terminator and
/// stripped (consistent with `str::lines()` semantics). To append a blank
/// line after the last inserted line, end `new_text` with `"\n\n"`.
```

And a matching test:
```rust
#[test]
fn insert_trailing_newline_in_text_stripped() {
    // "foo\n" and "foo" produce the same insert — trailing \n is a terminator.
    let content = "a\nb\n";
    assert_eq!(
        insert_lines_before(content, 1, "foo\n"),
        insert_lines_before(content, 1, "foo")
    );
}
```

---

### Round 3 delta: all resolved ✅

| Round 3 finding | Status |
|---|---|
| `insert_after` empty-text rejection untested | ✅ `insert_after_empty_text_rejected` added |
| Stale-anchor path for insert ops untested | ✅ `stale_anchor_before_insert` added |
| `text.lines().count() == 0` → `text.is_empty()` | ✅ Fixed |
| `text.lines().count()` computed twice | ✅ `text_line_count` bound once |
| `insert_empty_text_rejected` primed unused cache | ✅ Fixed, with explanatory comment |

---

### Round 4 summary

| Finding | Severity |
|---|---|
| `text_line_count` computed for Replace ops and on early-return path (dead work) | Should fix |
| Trailing `\n` in `text` silently stripped — undocumented and untested | Should fix |

decko (commented):
Both findings addressed in commit 7c6f3da:

**Finding 1 — dead `text_line_count` computation** ✅ Removed the unconditional `let text_line_count = text.lines().count();` from the outer scope (where it was dead for Replace ops and discarded on the insert+empty early-return path). The count is now computed inline in each InsertBefore/InsertAfter arm of the success message — only runs when actually used.

**Finding 2 — trailing `\n` stripping undocumented** ✅ Added doc note to `insert_lines_before`: "A trailing `\n` in `new_text` is treated as a line terminator and stripped (consistent with `lines()` semantics). To append a blank line after the last inserted line, end `new_text` with `"\n\n"`." Added `insert_trailing_newline_in_text_stripped` test verifying `insert_lines_before(..., "foo\n")` == `insert_lines_before(..., "foo")`.

Test count: 213 → 214 (+1: `insert_trailing_newline_in_text_stripped`)
Verification: `cargo build` ✅ `cargo test` (214) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 5

All round 4 findings resolved. Docstring updated, `text_line_count` moved inside the write-success branch, test added. Two genuine issues remain.

---

### Finding 1: `insert_trailing_newline_in_text_stripped` test doesn't pin the actual value

The new test (lines 767–777) asserts:

```rust
assert_eq!(
    insert_lines_before(content, 1, "foo\n"),
    insert_lines_before(content, 1, "foo")
);
```

This only proves the two inputs produce *the same output*. It says nothing about what that output is. A broken `insert_lines_before` that returned `content` unchanged for both would pass this test.

The test should pin the concrete expected value:

```rust
#[test]
fn insert_trailing_newline_in_text_stripped() {
    let content = "a\nb\n";
    // trailing \n in text is a terminator, not content — same result as without it
    assert_eq!(insert_lines_before(content, 1, "foo\n"), "a\nfoo\nb\n");
    assert_eq!(insert_lines_before(content, 1, "foo"),   "a\nfoo\nb\n");
}
```

---

### Finding 2: `text` parameter description (LLM-facing contract) not updated for trailing-`\n` stripping

The `insert_lines_before` docstring was correctly updated (lines 320–323) to document that a trailing `\n` in `new_text` is treated as a terminator. But the `text` field's `description` in `parameters_schema()` — the string an LLM caller actually receives — still says only:

> "New text to insert or replace with. Include newline characters between lines as needed."

A caller that passes `"text": "fn foo() {\n}\n"` expecting to insert the function will get the right result, but a caller trying `"text": "fn foo() {\n}\n\n"` to append a trailing blank line won't know why they need the double newline. The behaviour is documented internally but not at the API boundary.

Suggested update to the `text` description:

```
"New text to insert or replace with. Include newline characters between lines as needed. \
 For insert operations, a trailing newline is treated as a line terminator and stripped; \
 to also insert a trailing blank line, end with '\\n\\n'."
```

---

### Finding 3 (nit): `InsertBefore`/`InsertAfter` success-message arms duplicate `text.lines().count()`

Lines 281–296 have two near-identical arms:

```rust
EditOp::InsertBefore => format!("Inserted {} line(s) before anchor '{}' ...", text.lines().count(), ...),
EditOp::InsertAfter  => format!("Inserted {} line(s) after anchor '{}' ...",  text.lines().count(), ...),
```

Only "before"/"after" differs. They can be merged to eliminate the duplication and make both variants obviously consistent:

```rust
EditOp::InsertBefore | EditOp::InsertAfter => {
    let dir = if op == EditOp::InsertBefore { "before" } else { "after" };
    format!("Inserted {} line(s) {} anchor '{}' ({}→{} bytes) in {}",
        text.lines().count(), dir, anchor, old_bytes, new_bytes, resolved.display())
}
```

---

### Finding 4 (nit): Documented single-line + empty-replace edge case has no test

The comment at lines 371–376 explicitly documents that replacing the only line of a single-line file with `""` collapses the file to empty — a potentially surprising departure from the multi-line "blank the line" semantic. Verified: `replace_line_range("only\n", 0, 0, "")` → `""`.

Since the comment calls it out as a known edge case, a test pinning this behaviour would prevent a future refactor from silently changing it:

```rust
#[test]
fn replace_single_line_with_empty_collapses_to_empty_file() {
    // Known edge case: sole line + empty replacement → empty file, not blank line.
    assert_eq!(replace_line_range("only\n", 0, 0, ""), "");
}
```

---

### Round 4 delta: all resolved ✅

| Round 4 finding | Status |
|---|---|
| `text_line_count` computed unconditionally for Replace | ✅ Binding removed; inline in insert arms |
| Trailing `\n` in `text` undocumented and untested | ✅ Docstring updated + `insert_trailing_newline_in_text_stripped` added |

### Round 5 summary

| Finding | Severity |
|---|---|
| `insert_trailing_newline_in_text_stripped` only checks equality, not value | Should fix |
| `text` parameter description not updated (LLM-facing contract) | Should fix |
| `InsertBefore`/`InsertAfter` success arms duplicate `text.lines().count()` | Nit |
| Single-line + empty-replace edge case commented but untested | Nit |

decko (commented):
All four findings addressed in commit f85aad0:

**Finding 1 — test pins values** ✅ `insert_trailing_newline_in_text_stripped` now asserts against concrete expected strings: both `"foo\n"` and `"foo"` produce `"a\nfoo\nb\n"`. A broken no-op implementation can no longer pass.

**Finding 2 — text parameter description** ✅ Updated the `text` field description in `parameters_schema()`: "For insert operations, a trailing newline is treated as a line terminator and stripped; to also insert a trailing blank line, end with \n\n." The LLM-facing contract now matches the internal docstring.

**Finding 3 — merge success arms** ✅ Combined `InsertBefore`/`InsertAfter` arms into a single `EditOp::InsertBefore | EditOp::InsertAfter` arm with a dynamic `dir` variable ("before"/"after"). Eliminates the near-duplicate `format!` calls.

**Finding 4 — edge case test** ✅ Added `replace_single_line_with_empty_collapses_file`: `replace_line_range("only\n", 0, 0, "") → ""`. Pins the commented edge case where a single-line file collapses to empty.

Test count: 214 → 215 (+1: `replace_single_line_with_empty_collapses_file`)
Verification: `cargo build` ✅ `cargo test` (215) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 6

All round 5 findings resolved cleanly. Two new issues on this pass, both verified with `rustc`.

---

### Finding 1: `&& new_text.is_empty()` in `replace_line_range` is a redundant condition

Line 377:
```rust
if new_lines.is_empty() && new_text.is_empty() {
    new_lines.push("");
}
```

`new_lines` is collected from `new_text.lines()`. As confirmed by exhaustive testing across all plausible inputs, there is no non-empty string whose `lines()` iterator yields zero items — `new_lines.is_empty()` is true if and only if `new_text.is_empty()`. The conjunction reduces to `new_lines.is_empty()` alone.

The redundant half looks like it's guarding against a case that can't happen (`new_text` non-empty but producing no lines). A reader unfamiliar with `str::lines()` semantics might conclude such a case exists, or wonder why the extra check was needed.

```rust
// simpler and equally correct:
if new_lines.is_empty() {
    new_lines.push("");
}
```

---

### Finding 2: `text` parameter description documents trailing-`\n` stripping for insert only, but it applies to `replace` too

The updated description on line 82 reads:

> "…For insert operations, a trailing newline is treated as a line terminator and stripped; to also insert a trailing blank line, end with '\\n\\n'."

But `replace_line_range` also collects `new_text.lines()`, so `"X\n"` and `"X"` produce identical replacements. Verified: `replace_line_range("a\nb\nc\n", 1, 1, "X\n")` equals `replace_line_range("a\nb\nc\n", 1, 1, "X")`.

The current wording implies this behaviour is specific to insert, when it is universal. A caller passing `"text": "fn foo() {\n}\n"` to a `replace` operation has the same silent stripping. Suggested revision:

```
"New text to insert or replace with. Use '\\n' to separate lines. A trailing newline \
 is treated as a line terminator and stripped for all operations. For insert operations, \
 to also append a trailing blank line, end with '\\n\\n'."
```

---

### Round 5 delta: all resolved ✅

| Round 5 finding | Status |
|---|---|
| `insert_trailing_newline_in_text_stripped` only checked equality, not value | ✅ Both assertions now pin concrete expected string |
| `text` description not updated (LLM contract) | ✅ Description updated with stripping note |
| `InsertBefore`/`InsertAfter` arms duplicated `text.lines().count()` | ✅ Arms merged, single call |
| Single-line + empty-replace collapses to `""` — undocumented | ✅ `replace_single_line_with_empty_collapses_file` test added |

### Round 6 summary

| Finding | Severity |
|---|---|
| `&& new_text.is_empty()` is redundant and misleading in `replace_line_range` | Should fix |
| `text` description attributes trailing-`\n` stripping to insert ops only | Should fix |

decko (commented):
Both findings addressed in commit 072b45c:

**Finding 1 — redundant condition** ✅ Removed `&& new_text.is_empty()` from `replace_line_range`. `new_lines.is_empty()` is true iff `new_text` is empty — no non-empty string produces zero lines via `str::lines()`. The second condition was misleading. Simplified to `if new_lines.is_empty()` with updated comment documenting the equivalence.

**Finding 2 — trailing-\\n stripping in `text` description** ✅ Updated from "For insert operations, ..." to "for all operations." Both insert and replace use `str::lines()`, so both strip trailing newlines identically. The `\\n\\n` guidance for trailing blank lines remains specific to insert (replace has no "blank line" semantic).

Test count: 215 (unchanged — comment-only change)
Verification: `cargo build` ✅ `cargo test` (215) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 7

Both round 6 findings resolved. The redundant condition is gone (with an explanatory comment), and the `text` description now correctly says "stripped for all operations." One substantive issue remains.

---

### Finding 1: Trailing-`\n` stripping in `new_text` is documented for all operations but only tested for insert

The `text` parameter description (line 82) now reads:

> "A trailing newline is treated as a line terminator and stripped **for all operations**."

`insert_trailing_newline_in_text_stripped` (lines 776–784) verifies this for `insert_lines_before`. But `replace_line_range` also collects `new_text.lines()` and has the same stripping behaviour — and there is no test confirming it.

The claim "for all operations" in the description is now unsupported by the test suite. A symmetrical test for replace:

```rust
#[test]
fn replace_trailing_newline_in_text_stripped() {
    // str::lines() strips a trailing \n — same as for insert operations.
    let content = "a\nb\nc\n";
    assert_eq!(replace_line_range(content, 1, 1, "X\n"), "a\nX\nc\n");
    assert_eq!(replace_line_range(content, 1, 1, "X"),   "a\nX\nc\n");
}
```

---

### Nit: `\n\n` tip in `text` description is framed as insert-only but applies to replace too

Line 82 ends with: "For insert operations, to also append a trailing blank line, end with `\n\n`."

The same applies to replace — `"content\n\n"` in a replace produces the replaced range plus a trailing blank line. The tip is accurate for inserts but the "For insert operations" qualifier implies replace behaves differently. Removing the qualifier or extending it to "For all operations" would be more precise:

> "…To also produce a trailing blank line in the output, end with `'\\n\\n'`."

---

### Round 6 delta: all resolved ✅

| Round 6 finding | Status |
|---|---|
| `&& new_text.is_empty()` redundant in `replace_line_range` | ✅ Removed; comment explains the equivalence |
| `text` description scoped trailing-`\n` stripping to insert only | ✅ Updated to "for all operations" |

### Round 7 summary

| Finding | Severity |
|---|---|
| `replace_line_range` has no test for trailing-`\n` stripping in `new_text` | Should fix |
| `\n\n` tip framed as insert-specific in description | Nit |

decko (commented):
## Rust Specialist Review — Round 8

No new commits since round 7 — the two round 7 findings (missing `replace_trailing_newline_in_text_stripped` test; `\n\n` tip framed as insert-specific) remain open. Independent fresh pass found two additional issues.

---

### Finding 1: `replace_line_range` docstring is missing the `new_text` semantics note present in `insert_lines_before`

`insert_lines_before` (line 313) has an explicit Note block documenting:
- CRLF normalisation (`\r\n` → `\n`)
- Trailing `\n` in `new_text` stripped
- How to append a blank line (`\n\n`)

`replace_line_range` (line 358) has none of these. Both functions use the same `new_text.lines().collect()` mechanism with identical stripping behaviour. The asymmetry means a developer reading `replace_line_range`'s docstring in isolation gets an incomplete picture of how `new_text` is processed.

Suggested addition to `replace_line_range`'s docstring:

```rust
/// Note: `str::lines()` normalizes `\r\n` → `\n`, so Windows-style line
/// endings are converted to LF on write. A trailing `\n` in `new_text` is
/// treated as a line terminator and stripped; `"X\n"` and `"X"` produce
/// the same replacement.
```

---

### Finding 2: `EditOp` derives `PartialEq` but not `Eq`

Line 20:
```rust
#[derive(Copy, Clone, Debug, PartialEq)]
enum EditOp { Replace, InsertBefore, InsertAfter }
```

`EditOp` contains no floats and has reflexive equality — it satisfies `Eq`'s contract. The idiomatic Rust convention is to derive `Eq` alongside `PartialEq` whenever both hold. Omitting `Eq` is flagged by the `clippy::nursery::derive_partial_eq_without_eq` lint (not in the default `clippy::all` set, so `-D warnings` doesn't catch it, but it's a real gap).

Practical impact here is low since `EditOp` is private and never used in collections. But it's a one-character fix that makes the type contract complete:

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```

---

### Round 7 findings (still open)

| Finding | Status |
|---|---|
| Missing `replace_trailing_newline_in_text_stripped` test | ⏳ Open |
| `\n\n` tip in `text` description framed as insert-only | ⏳ Open (nit) |

### Round 8 summary

| Finding | Severity |
|---|---|
| `replace_line_range` docstring missing `new_text` semantics (CRLF, trailing-`\n` stripping) | Should fix |
| `EditOp` derives `PartialEq` without `Eq` | Nit |

decko (commented):
All four findings from rounds 7 and 8 addressed in commit 50bdd09:

**Round 7:**
- **F1 — missing replace trailing-\\n test** ✅ Added `replace_trailing_newline_in_text_stripped`. Verifies `replace_line_range("a\\nb\\nc\\n", 1, 1, "X\\n")` == `replace_line_range(..., "X")` == `"a\\nX\\nc\\n"`. The "all operations" claim in the text description is now backed by tests for both insert and replace.
- **Nit — \\n\\n tip insert-only** ✅ Removed "For insert operations" qualifier. Now reads: "To also produce a trailing blank line in the output, end with \n\n."

**Round 8:**
- **F1 — replace_line_range docstring parity** ✅ Added `new_text` semantics note (CRLF normalization, trailing-\\n stripping) mirroring the existing note in `insert_lines_before`.
- **F2 — EditOp missing Eq** ✅ Added `Eq` to the derive list on `EditOp`.

Test count: 215 → 216 (+1: `replace_trailing_newline_in_text_stripped`)
Verification: `cargo build` ✅ `cargo test` (216) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — Round 9

All four open findings from rounds 7–8 resolved: `Eq` derived, `replace_line_range` docstring updated with Note block, `replace_trailing_newline_in_text_stripped` test added, `\n\n` tip de-scoped from insert-only. One finding on this pass.

---

### Finding 1: The `\n\n` → trailing-blank-line behaviour is documented but untested

Both the `text` parameter description (line 82) and the `insert_lines_before` docstring (lines 322–323) promise:

> "To also produce a trailing blank line in the output, end with `\n\n`."

`insert_trailing_newline_in_text_stripped` (line 789) tests the single-trailing-`\n` strip, and `replace_trailing_newline_in_text_stripped` (line 495) mirrors it for replace. But neither tests the `\n\n` case. Verified with `rustc`:

```
insert_lines_before("a\nb\n", 1, "foo\n\n")  → "a\nfoo\n\nb\n"  ✓
replace_line_range("a\nb\nc\n", 1, 1, "X\n\n") → "a\nX\n\nc\n"  ✓
```

The behaviour is correct, but the documentation makes a specific promise that no test backs. Suggested additions:

```rust
#[test]
fn insert_double_newline_appends_blank_line() {
    // "\n\n" keeps one line of content + appends a blank line.
    let content = "a\nb\n";
    assert_eq!(insert_lines_before(content, 1, "foo\n\n"), "a\nfoo\n\nb\n");
}

#[test]
fn replace_double_newline_appends_blank_line() {
    let content = "a\nb\nc\n";
    assert_eq!(replace_line_range(content, 1, 1, "X\n\n"), "a\nX\n\nc\n");
}
```

---

### Round 8 delta: all resolved ✅

| Round 8 finding | Status |
|---|---|
| `replace_line_range` docstring missing `new_text` semantics | ✅ Note block added |
| `EditOp` derives `PartialEq` without `Eq` | ✅ `Eq` added |
| Missing `replace_trailing_newline_in_text_stripped` test (round 7) | ✅ Added |
| `\n\n` tip framed as insert-specific (round 7 nit) | ✅ De-scoped to "all operations" |

### Round 9 summary

| Finding | Severity |
|---|---|
| `\n\n` trailing-blank-line behaviour documented but untested | Should fix |

decko (commented):
Round 9 finding addressed in commit c2750e8:

**Finding 1 — \\n\\n trailing-blank-line documented but untested** ✅
Added two symmetric tests pinning the `"\n\n"` behaviour:
- `insert_double_newline_appends_blank_line`: `insert_lines_before("a\nb\n", 1, "foo\n\n")` → `"a\nfoo\n\nb\n"`
- `replace_double_newline_appends_blank_line`: `replace_line_range("a\nb\nc\n", 1, 1, "X\n\n")` → `"a\nX\n\nc\n"`
Both confirm one content line + trailing blank line are preserved.

Test count: 216 → 218 (+2)
Verification: `cargo build` ✅ `cargo test` (218) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

decko (commented):
## Rust Specialist Review — PR #83: `insert_before` / `insert_after` operations

### Overview

Clean addition of two new `edit_file` operations. The delegation pattern (`insert_after → insert_lines_before(at+1)`), the `EditOp` enum, and the backward-compatible defaulting to `replace` are all solid design choices. Test coverage is comprehensive. Below are the findings by severity.

---

### Issues

#### 1. Success message: `if op ==` inside an already-matched arm (minor style)

In the `Ok(())` branch, the `InsertBefore | InsertAfter` arm re-compares `op` with an `if`:

```rust
EditOp::InsertBefore | EditOp::InsertAfter => {
    let dir = if op == EditOp::InsertBefore { "before" } else { "after" };
    ...
}
```

The compiler can't prove exhaustiveness here and the `else` silently covers `InsertAfter`. A nested `match` is clearer and will catch a missing arm if a new variant is ever added:

```rust
EditOp::InsertBefore => format!("Inserted {} line(s) before anchor '{}' ...", ...),
EditOp::InsertAfter  => format!("Inserted {} line(s) after anchor '{}' ...", ...),
```

#### 2. `end_line` dummy value is a subtle code smell (minor)

When `op != Replace`, the anchor-resolution block returns `(start, start)` with the comment `// end_line unused for insert ops`. This means `end_line` always has a value but is conditionally meaningless. An `Option<usize>` would make the invariant explicit at the type level and eliminate the ambiguity:

```rust
let end_line: Option<usize> = if op == EditOp::Replace { Some(end) } else { None };
```

#### 3. `from_str` as an inherent method instead of `FromStr` trait (nit)

`EditOp::from_str` returns `Option<Self>`, which is fine for a private internal type. However, implementing `std::str::FromStr` (returning `Result`) is the Rust convention and would make it usable with `.parse::<EditOp>()` and better tooling integration. For a private enum this is a low-priority nit, but worth considering if `EditOp` ever becomes part of a public interface.

#### 4. `text.lines().count()` double-splits in the success message (nit)

The success message counts inserted lines with `text.lines().count()`. The string was already split inside `insert_lines_before`. This is completely negligible at runtime, but if the helper functions ever returned the line count as part of their result, this redundant split would disappear naturally.

---

### What's Done Well

- **`insert_after` delegation** to `insert_lines_before(at.saturating_add(1))` is elegant and avoids duplication.
- **`debug_assert!` in `insert_lines_before`** catches out-of-bounds `at` in debug builds while the `start_line >= line_count` guard covers release builds — both layers are correct.
- **`replace_line_range` simplification**: removing the redundant `&& new_text.is_empty()` from the old condition is correct — `new_lines.is_empty()` is already true iff `new_text == ""` (Rust's `str::lines()` only returns an empty iterator for an empty string).
- **CRLF normalization** is documented consistently across all helpers — callers won't be surprised.
- **JSON schema** correctly omits `operation` from `required` and documents `end_anchor` as required only for `replace` — the runtime validation matches the documentation.
- **Empty-text guard** for insert ops with a clear error message is a good UX choice. Distinguishing `insert nothing` (mistake) from `replace with empty` (blank the line) is a meaningful semantic difference.
- **Test coverage** is thorough: stale-anchor detection, cache invalidation, CRLF normalization, trailing-newline edge cases, and both empty-text variants all have dedicated tests.

---

### Summary

This is a solid PR. The two actionable items are the nested-`match` in the success message and the `Option<usize>` for `end_line` — both improve readability and future maintainability. Everything else is nits. The core logic, error handling, and tests are correct.

decko (commented):
Round 10 actionable items addressed in commit bc2df91:

**1. if op == inside matched arm** ✅ Split the combined `InsertBefore | InsertAfter` arm back into two separate arms (`InsertBefore` and `InsertAfter`). No `if op ==` needed — each arm directly formats its own message. The compiler can prove exhaustiveness.

**2. end_line dummy value** ✅ Changed `end_line` from `usize` to `Option<usize>`: `Some(end)` for Replace, `None` for insert ops. The invariant is explicit at the type level. Guard uses `if let Some(end) = end_line`; Replace arms use `.unwrap()` — guaranteed safe since Replace always sets the value.

Nits 3-4 (FromStr trait, text.lines() double-split) are acknowledged but not addressed — they are future-facing concerns, not current defects.

Test count: 218 (unchanged — pure refactor)
Verification: `cargo build` ✅ `cargo test` (218) ✅ `cargo clippy -D warnings` ✅ `cargo fmt -- --check` ✅

PR #94, opened 2026-05-26
feat(edit): multi-file files array schema + per-file execute loop (#89)

### LINKED ISSUE #89: F3d: multi-file files array schema + per-file loop in execute()
## Context
Sub-task of #29 (F3: multi-file batching). Builds on F3a, F3b, F3c.

**Parent:** #29
**Depends on:** F3a, F3b, F3c (helpers, enum, overlaps must exist)

## Scope
- Replace single-file `path`/`anchor`/`end_anchor`/`text` params with `files` array schema
- New schema: `{ files: [{ path, edits: [{ edit_type, anchor, end_anchor?, text }] }] }`
- Per-file loop in `execute()`: resolve path, resolve all anchors, check overlaps, read file, apply all edits, write back, invalidate cache
- Call `check_overlaps` within each file's edit batch
- Multi-file result formatting: one line per file with byte delta
- Integration test: 2 files modified in one call

**Bottom-to-top ordering** is handled in F3e.

**Type:** 🟠 DENSE (~100 lines)

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
F3d: Replace single-file edit schema with multi-file `files` array. Per-file loop in execute() with batch anchor resolution, overlap checking, and partial-success semantics.

## Changes
- **Schema**: `{ files: [{ path, edits: [{ operation?, anchor, end_anchor?, text }] }] }`
- **execute() rewrite**: validate edits → resolve path → resolve anchors (batch) → check_overlaps → read file → guard line counts → apply edits bottom-to-top → write back → invalidate cache
- **Bottom-to-top ordering**: edits sorted by `start` descending before application (preserves line indices after insertions/replacements)
- **Partial success**: `ToolResult::ok()` if any file succeeded, `ToolResult::error()` only if all files failed
- **No panics**: `.expect()` replaced with `match` on lock poisoning
- **15 integration tests** updated + 3 new multi-file tests (happy path, partial success, all fail)
- `single_file_edit()` test helper reduces boilerplate

## Design decisions
- All edits for a file are validated BEFORE any I/O (fails fast on bad params)
- Anchor resolution in batch within one lock acquisition
- `check_overlaps()` rejects overlapping edits within a file
- Cache invalidated after successful write; lock poisoning is non-fatal on cache invalidation
- Result format: `Applied N edit(s) (X→Y bytes) to path` per file

## Verification
- [x] cargo build ✅
- [x] cargo test (227/227) ✅  
- [x] cargo clippy -- -D warnings ✅
- [x] cargo fmt ✅
- [x] rust-reviewer findings addressed ✅

Closes #89

### DIFF
--- a/src/tools/edit.rs
+++ b/src/tools/edit.rs
@@ -212,6 +212,7 @@ impl Tool for EditFileTool {
                     };
 
                     let text = match edit_entry.get("text").and_then(Value::as_str) {
+                        // Replace allows empty text ("blank the line"); inserts require content.
                         Some(t) if !t.is_empty() || op == EditOp::Replace => t,
                         Some(_) => {
                             edit_error = Some(format!(
@@ -262,8 +263,16 @@ impl Tool for EditFileTool {
                 // line count are not detected. This is inherent to the caching
                 // design and acceptable for this scope.
                 let edits: Vec<ResolvedEdit> = {
-                    let mut anchor_state =
-                        ctx.anchor_state.lock().expect("anchor state lock poisoned");
+                    let mut anchor_state = match ctx.anchor_state.lock() {
+                        Ok(guard) => guard,
+                        Err(_) => {
+                            results.push(format!(
+                                "edit_file: anchor state lock poisoned for '{}'",
+                                path_str
+                            ));
+                            continue;
+                        }
+                    };
                     let anchors = match anchor_state.get_anchors(&resolved) {
                         Ok(a) => a,
                         Err(e) => {
@@ -393,10 +402,15 @@ impl Tool for EditFileTool {
                     continue;
                 }
 
-                // --- apply edits in declaration order ---
+                // --- apply edits bottom-to-top (preserves line indices) ---
                 let old_bytes = content.len();
                 let mut new_content = content;
-                for edit in &edits {
+                // Sort descending by start: higher lines applied first so
+                // lower-line edits' positions remain valid after insertions
+                // or multi-line replacements above them.
+                let mut sorted: Vec<&ResolvedEdit> = edits.iter().collect();
+                sorted.sort_by_key(|e| std::cmp::Reverse(e.start));
+                for edit in &sorted {
                     new_content = match edit.operation {
                         EditOp::Replace => {
                             replace_line_range(&new_content, edit.start, edit.end, &edit.text)
@@ -417,9 +431,15 @@ impl Tool for EditFileTool {
 
                         // Invalidate anchor cache.
                         {
-                            let mut anchor_state =
-                                ctx.anchor_state.lock().expect("anchor state lock poisoned");
-                            anchor_state.notify_edit(&resolved);
+                            match ctx.anchor_state.lock() {
+                                Ok(mut anchor_state) => {
+                                    anchor_state.notify_edit(&resolved);
+                                }
+                                Err(_) => {
+                                    // Cache invalidation failure is non-fatal;
+                                    // the file was already written successfully.
+                                }
+                            }
                         }
 
                         results.push(format!(
@@ -438,9 +458,7 @@ impl Tool for EditFileTool {
             }
 
             // --- assemble final result ---
-            if results.is_empty() {
-                Ok(ToolResult::error("edit_file: no files processed"))
-            } else if any_success {
+            if any_success {
                 Ok(ToolResult::ok(results.join("\n")))
             } else {
                 Ok(ToolResult::error(results.join("\n")))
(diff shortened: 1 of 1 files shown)


src/tools/edit.rs
```diff
@@ -2,7 +2,9 @@
 //!
 //! Uses stable word anchors (from [`read_file`]) to target edit locations.
 //! Supports three operations: `replace` (the default), `insert_before`,
-//! and `insert_after`. Multi-file batching is implemented in a subsequent PR.
+//! and `insert_after`. Edits are batched per file via `check_overlaps`; edits
+//! within a file are applied in declaration order (bottom-to-top sorting is
+//! handled in a follow-up PR).
```
decko: **[Major]** This comment is stale — bottom-to-top sorting via `sorted.sort_by_key(|e| std::cmp::Reverse(e.start))` is implemented in THIS PR (line 412). Update to reflect reality, e.g.: `Edits within a file are applied bottom-to-top (higher start lines first) to preserve line indices after insertions.`

src/tools/edit.rs
```diff
@@ -408,6 +565,42 @@ fn replace_line_range(content: &str, start_line: usize, end_line: usize, new_tex
...
+/// `a.start <= b.end && b.start <= a.end`.
+///
+/// Adjacent inserts (e.g., insert_after at line 1 and insert_before at line 2)
+/// are NOT overlapping — their ranges `[1,1]` and `[2,2]` are disjoint.
+///
+/// # Errors
+///
+/// Returns `Err(ToolResult::error(...))` with a message describing the first
+/// overlap found, or `Ok(())` if no overlaps are detected.
+fn check_overlaps(edits: &[ResolvedEdit], path: &str) -> Result<(), ToolResult> {
```
decko, line 583: **[Major]** Returning `Result<(), ToolResult>` leaks the tool-layer type into a pure utility function. Prefer `Result<(), String>` and let the caller wrap the error — currently the call site immediately extracts `.content` anyway (`results.push(err.content)`). Separately: the overlap predicate is a false positive for `insert_after(N)` + `insert_before(N)` — both produce range `[N,N]`, satisfy `a.start <= b.end && b.start <= a.end`, and are wrongly rejected. A "wrap this line" use-case (insert header before + trailer after the same anchor) breaks silently. Consider exempting same-line point-inserts from the overlap check.

src/tools/edit.rs
```diff
@@ -92,219 +127,341 @@ impl Tool for EditFileTool {
...
-                            "Inserted {} line(s) after anchor '{}' ({}→{} bytes) in {}",
-                            text.lines().count(),
-                            anchor,
+                // --- apply edits bottom-to-top (preserves line indices) ---
+                let old_bytes = content.len();
+                let mut new_content = content;
+                // Sort descending by start: higher lines applied first so
+                // lower-line edits' positions remain valid after insertions
+                // or multi-line replacements above them.
+                let mut sorted: Vec<&ResolvedEdit> = edits.iter().collect();
```
decko: **[Major]** Unnecessary allocation: collecting `edits.iter()` into a `Vec<&ResolvedEdit>` just to sort it. `edits` is not borrowed elsewhere after this point — declare it `mut` earlier and call `edits.sort_by_key(|e| std::cmp::Reverse(e.start));` in place. `edits.len()` in the success message below still works.

decko (commented):
## Rust Specialist Review — Major findings only

**Overall:** The multi-file refactor is structurally sound — no panics, correct lock discipline, bottom-to-top ordering, and partial-success semantics are all well-implemented. Three issues need attention before merge.

### Blocking
None.

### Major

**1. Module doc comment contradicts the implementation (line 6-7)**
The doc says bottom-to-top sorting is "handled in a follow-up PR" but `sorted.sort_by_key(|e| std::cmp::Reverse(e.start))` is right here in this PR. Stale doc will mislead every future reader of the module banner.

**2. `check_overlaps` false-positive: `insert_after(N)` + `insert_before(N)` at the same line (line 580)**
Both edits get `start=N, end=N`. The predicate `a.start <= b.end && b.start <= a.end` fires, rejecting a semantically valid "wrap this line" pattern. The test suite covers `insert_after(1)` + `insert_before(2)` (adjacent, disjoint) but not `insert_after(N)` + `insert_before(N)` (same-line, should be allowed). Fix: when both edits are point-inserts (`end == start`) on the same line, check the operation pair before declaring overlap.

**3. Unnecessary allocation in the apply loop (lines 411-412)**
`sorted: Vec<&ResolvedEdit>` allocates a secondary reference-vec just to avoid mutating `edits`. Since `edits` is never passed anywhere after this point (only `edits.len()` is needed), just declare it `mut` and sort in place. The borrow on `edits.len()` after the write is fine even with an in-place sort.

decko (commented):
## Rust Specialist Follow-up Review

All three major findings from the previous round are resolved:

- ✅ **Module doc** — now correctly describes bottom-to-top ordering
- ✅ **Same-line wrap false positive** — `is_same_line_point_ops && is_wrap_pair` exception added, test `check_overlaps_same_line_wrap_ok` covers it
- ✅ **Unnecessary allocation** — `edits` is now `let mut`, sorted in place, secondary `Vec<&ResolvedEdit>` gone

**One minor carry-over (not blocking):** `check_overlaps` still returns `Result<(), ToolResult>`. The call site immediately extracts `.content` (`results.push(err.content)`), meaning the tool-layer type leaks into a pure geometry function. `Result<(), String>` would be cleaner and easier to test in isolation. Fine to leave for a follow-up.

**No blocking or major issues remain. LGTM.**

PR #97, opened 2026-05-26
feat(treesitter): language loading + grammar registry (#30)

### LINKED ISSUE #30: G1: Tree-sitter language loading + grammar registry
## Context
`src/treesitter/languages.rs`: Load tree-sitter grammars for Rust, Python, TypeScript and map file extensions to languages.

**⚠️ Blocked by:** Issue #4 (resolve tree-sitter version compatibility)

## Scope
- Language enum (Rust, Python, TypeScript)
- Grammar loading via `tree-sitter-{rust,python,typescript}` crates
- `language_for_path(path)` → extension → language
- Integration with `build.rs` if C compilation needed

**Test:** Unit: .rs→Rust, .py→Python, .ts→TypeScript

## DoD
- [ ] cargo build (with tree-sitter deps from #4)
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
G1: Tree-sitter language registry. Map file extensions to languages and load grammar definitions from the three grammar crates.

## Implementation
- **Language enum**: Rust, Python, TypeScript, Tsx
- **language_for_path(path)**: extension → Language mapping (.rs, .py, .ts, .tsx)
- **language_grammar(lang)**: loads tree_sitter::Language via grammar crate constants (LanguageFn → Into<Language>)
- **8 unit tests**: extension mapping (6) + grammar loading (2)

## Grammar API
Each grammar crate exposes a const LanguageFn:
- `tree_sitter_rust::LANGUAGE`
- `tree_sitter_python::LANGUAGE`
- `tree_sitter_typescript::LANGUAGE_TYPESCRIPT` / `LANGUAGE_TSX`

LanguageFn implements Into<tree_sitter::Language> via the tree-sitter-language crate.

## Verification
- [x] cargo build ✅
- [x] cargo test (236/236) ✅
- [x] cargo clippy ✅
- [x] cargo fmt ✅

Closes #30

### DIFF
--- a/src/treesitter/mod.rs
+++ b/src/treesitter/mod.rs
@@ -17,7 +17,7 @@ pub enum Language {
     Rust,
     /// Python (`.py`)
     Python,
-    /// TypeScript (`.ts`)
+    /// TypeScript (`.ts`, `.js`)
     TypeScript,
     /// TSX / React TypeScript (`.tsx`)
     Tsx,
@@ -46,9 +46,7 @@ pub fn language_grammar(lang: Language) -> tree_sitter::Language {
 
 /// Map a file path to its tree-sitter [`Language`].
 ///
-/// Returns `None` for unsupported or unrecognized extensions. The path
-/// must contain a file extension — bare extension strings like `"rs"`
-/// are not matched (use a full or relative path like `"file.rs"`).
+/// Returns `None` for unsupported or unrecognized extensions.
 ///
 /// # Examples
 ///
@@ -57,15 +55,16 @@ pub fn language_grammar(lang: Language) -> tree_sitter::Language {
 /// assert_eq!(language_for_path("src/main.rs"), Some(Language::Rust));
 /// assert_eq!(language_for_path("script.py"), Some(Language::Python));
 /// assert_eq!(language_for_path("index.ts"), Some(Language::TypeScript));
+/// assert_eq!(language_for_path("app.js"), Some(Language::TypeScript));
 /// assert_eq!(language_for_path("component.tsx"), Some(Language::Tsx));
 /// assert_eq!(language_for_path("Makefile"), None);
 /// ```
-pub fn language_for_path(path: &str) -> Option<Language> {
-    let ext = Path::new(path).extension()?.to_str()?;
+pub fn language_for_path(path: impl AsRef<Path>) -> Option<Language> {
+    let ext = path.as_ref().extension()?.to_str()?;
     match ext {
         "rs" => Some(Language::Rust),
         "py" => Some(Language::Python),
-        "ts" => Some(Language::TypeScript),
+        "ts" | "js" => Some(Language::TypeScript),
         "tsx" => Some(Language::Tsx),
         _ => None,
     }
@@ -85,6 +84,11 @@ mod tests {
     fn rust_extension() {
         assert_eq!(language_for_path("main.rs"), Some(Language::Rust));
         assert_eq!(language_for_path("src/lib.rs"), Some(Language::Rust));
+        // AsRef<Path> accepts PathBuf
+        assert_eq!(
+            language_for_path(std::path::PathBuf::from("mod.rs")),
+            Some(Language::Rust)
+        );
     }
 
     #[test]
@@ -96,6 +100,7 @@ mod tests {
     #[test]
     fn typescript_extension() {
         assert_eq!(language_for_path("index.ts"), Some(Language::TypeScript));
+        assert_eq!(language_for_path("lib/util.js"), Some(Language::TypeScript));
     }
 
     #[test]
--- a/src/treesitter/mod.rs
+++ b/src/treesitter/mod.rs
@@ -1,1 +1,146 @@
-// C grammar bindings, .scm query files, parse tree caching.
+//! Tree-sitter language registry, grammar loading, and file-extension mapping.
+//!
+//! Grammar crates compile their own C source and expose language definitions
+//! via `tree_sitter_language::LanguageFn` constants. The module maps file
+//! extensions to [`Language`] variants and loads the corresponding grammar.
+
+use std::path::Path;
+
+// ---------------------------------------------------------------------------
+// Language enum
+// ---------------------------------------------------------------------------
+
+/// Supported tree-sitter languages.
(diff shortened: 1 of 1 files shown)


src/treesitter/mod.rs
```diff
@@ -1 +1,154 @@
...
+/// assert_eq!(language_for_path("app.js"), Some(Language::TypeScript));
+/// assert_eq!(language_for_path("component.tsx"), Some(Language::Tsx));
+/// assert_eq!(language_for_path("Makefile"), None);
+/// ```
+pub fn language_for_path(path: impl AsRef<Path>) -> Option<Language> {
+    let ext = path.as_ref().extension()?.to_str()?;
+    match ext {
+        "rs" => Some(Language::Rust),
+        "py" => Some(Language::Python),
+        "ts" | "js" => Some(Language::TypeScript),
```
decko, line 74: **Bug: `.js` routed to the TypeScript grammar**

`"ts" | "js" => Some(Language::TypeScript)` dispatches JavaScript files to `tree_sitter_typescript::LANGUAGE_TYPESCRIPT`. The TypeScript grammar is a superset of TypeScript, not JavaScript — it accepts TS-specific syntax that is invalid JS and models the AST with TS node types (e.g. `type_annotation`, `as_expression`). Any downstream query written against a JS file will receive a TS-grammar tree.

Concrete failure: a `.js` file containing a `with` statement (valid JS, invalid TS) will produce ERROR nodes in the parse tree, silently corrupting results for any structural query or anchor resolution.

Options:
1. Add `tree-sitter-javascript` crate + `Language::JavaScript` variant and route `.js` there.
2. Remove `.js` from this mapping and return `None` until a JS grammar is added.

```suggestion
        "ts" => Some(Language::TypeScript),
        "tsx" => Some(Language::Tsx),
```

src/treesitter/mod.rs
```diff
@@ -1 +1,154 @@
...
+/// # use carv::treesitter::{language_for_path, Language};
+/// assert_eq!(language_for_path("src/main.rs"), Some(Language::Rust));
+/// assert_eq!(language_for_path("script.py"), Some(Language::Python));
+/// assert_eq!(language_for_path("index.ts"), Some(Language::TypeScript));
+/// assert_eq!(language_for_path("app.js"), Some(Language::TypeScript));
+/// assert_eq!(language_for_path("component.tsx"), Some(Language::Tsx));
+/// assert_eq!(language_for_path("Makefile"), None);
+/// ```
+pub fn language_for_path(path: impl AsRef<Path>) -> Option<Language> {
+    let ext = path.as_ref().extension()?.to_str()?;
```
decko, line 65: **Extension matching is case-sensitive — breaks on macOS/Windows**

`path.as_ref().extension()?.to_str()?` yields the raw OS extension bytes. On case-insensitive filesystems (macOS HFS+, Windows NTFS) a file named `Main.RS` or `script.PY` is a valid Rust/Python file, but the match arms `"rs"` and `"py"` are byte-exact lowercase — `language_for_path` returns `None`.

Fix: lowercase before matching:

```suggestion
    let ext = path.as_ref().extension()?.to_str()?.to_ascii_lowercase();
    let ext = ext.as_str();
```

The `to_str()?` silently returning `None` for non-UTF-8 extensions is acceptable behavior but should be noted in the doc comment.

decko, line 65: **Dotfiles with language extensions return `None`**

`Path::new(".rs").extension()` returns `None` on all platforms — the OS treats the leading dot as the stem with no extension. A file named `.rs` or `.ts` (unusual but valid) gets no language mapping, silently falling through to `None` rather than being detected.

This is a hard limitation of `std::path::Path::extension()`. If the tool ever processes dotfile-named source files, the extension approach must be supplemented with a stem check. Worth documenting in the function doc comment so callers are not surprised.

src/treesitter/mod.rs
```diff
@@ -1 +1,154 @@
...
+
+    #[test]
+    fn empty_path() {
+        assert_eq!(language_for_path(""), None);
+    }
+
+    // -- language_grammar ----------------------------------------------------
+
+    #[test]
+    fn grammars_are_loadable() {
```
decko, line 140: **`grammars_are_loadable` only validates ABI compatibility, not parse correctness**

`set_language(&language)` succeeds if the grammar ABI version matches the `tree-sitter` core. But a grammar with a matching ABI version whose internal symbol/field tables are corrupt (e.g. stale build artifact, mislinked C object) would also pass this check — `set_language` does not parse any source.

Consider extending the test with a minimal parse to validate end-to-end grammar integrity:

```rust
let src = match variant {
    "Rust"       => "fn main() {}",
    "Python"     => "x = 1",
    "TypeScript" => "const x: number = 1;",
    "Tsx"        => "<div/>",
    _            => unreachable!(),
};
let tree = parser.parse(src, None)
    .unwrap_or_else(|| panic!("{variant} grammar produced no tree"));
assert!(!tree.root_node().has_error(), "{variant} grammar produced ERROR nodes");
```

decko (commented):
## Rust Specialist Review — PR #97 (language registry)

Overall the PR is clean: the grammar loading via `LanguageFn::into()` is idiomatic, the `language_for_path` signature (`impl AsRef<Path>`) is ergonomic and correct, and the 8 tests provide good baseline coverage.

### Findings summary (most severe first)

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **Bug** | `mod.rs:67` | `.js` files routed to the TypeScript grammar — wrong AST node types for JS-only constructs; `with` statements produce ERROR nodes |
| 2 | **Bug** | `mod.rs:63` | Extension matching is case-sensitive — `.RS`, `.PY` return `None` on macOS/Windows filesystems |
| 3 | **Design** | `mod.rs:63` | Dotfiles named `.rs`, `.ts` etc. return `None`; `Path::extension()` is a hard limit here — needs doc comment |
| 4 | **Test** | `mod.rs:126` | `grammars_are_loadable` only validates ABI compatibility; a minimal parse call per language would catch corrupt grammar linkage |

### Not flagged
- The `Language` enum not deriving `serde` traits is intentional omission (not a bug today); add when a consuming struct needs serialization.
- The deleted comment mentioned `.scm` query files and parse tree caching — these are understood to be follow-on work, not regressions.

PR #98, opened 2026-06-02
feat(treesitter): parser cache with invalidation hooks (#31)

### LINKED ISSUE #31: G2: Tree-sitter parser + parse tree cache + invalidation hooks
## Context
`src/treesitter/parser.rs`: Parse files into tree-sitter ASTs, cache results, invalidate on file modifications.

## Scope
- `parse_file(path)` → tree-sitter `Tree`
- Per-file cache: `HashMap<PathBuf, Tree>`
- Cache invalidation: triggered by `notify_edit()` from `AnchorState`
- Re-parse on next access after invalidation

**Type:** 🟠 Dense (~120 lines)

**Test:** Unit: parse fixture .rs file → valid tree; cache hit on re-parse; invalidate → re-parse

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
G2: Tree-sitter parser cache. Parse files into ASTs and cache results, with invalidation on file modifications.

## Implementation
- **ParserCache**: `HashMap\u003cPathBuf, tree_sitter::Tree\u003e` keyed by canonical path
- **parse_file(path)** → determines language from extension, reads file, parses, caches, returns cloned `Tree`
- **invalidate(path)** → removes cached tree (next parse_file re-reads and re-parses)
- **cache_size()** → count of cached entries (test introspection)
- Tree::clone() is cheap — pointer copy to C-heap data (Tree is Clone in tree-sitter 0.26.x)

## 6 unit tests
- Parse valid .rs file → no errors
- Parse valid .py file → no errors
- Cache hit returns identical tree structure
- Invalidate + reparse triggers fresh parse
- Unsupported extension → InvalidInput error
- File not found → NotFound error
- Cache size tracks add/invalidate

## Verification
- [x] cargo build ✅
- [x] cargo test (243/243) ✅
- [x] cargo clippy ✅
- [x] cargo fmt ✅

Closes #31

### DIFF
--- a/src/treesitter/mod.rs
+++ b/src/treesitter/mod.rs
@@ -4,6 +4,8 @@
 //! via `tree_sitter_language::LanguageFn` constants. The module maps file
 //! extensions to [`Language`] variants and loads the corresponding grammar.
 
+pub mod parser;
+
 use std::path::Path;
 
 // ---------------------------------------------------------------------------
--- a/src/treesitter/mod.rs
+++ b/src/treesitter/mod.rs
@@ -4,7 +4,7 @@
 //! via `tree_sitter_language::LanguageFn` constants. The module maps file
 //! extensions to [`Language`] variants and loads the corresponding grammar.
 
-pub mod parser;
+pub(crate) mod parser;
 
 use std::path::Path;
 
--- a/src/treesitter/parser.rs
+++ b/src/treesitter/parser.rs
@@ -1,0 +1,204 @@
+//! Tree-sitter parser cache with invalidation hooks.
+//!
+//! [`ParserCache`] manages per-file [`tree_sitter::Tree`] instances, avoiding
+//! redundant parses. The cache is invalidated when tools modify files (via
+//! `AnchorState::notify_edit` or direct calls to [`ParserCache::invalidate`]).
+
+use std::collections::HashMap;
+use std::io;
+use std::path::{Path, PathBuf};
+
+use super::{language_for_path, language_grammar};
+
+// ---------------------------------------------------------------------------
+// ParserCache
+// ---------------------------------------------------------------------------
+
+/// Per-file cache of parsed tree-sitter ASTs.
+///
+/// On first access, the file is read from disk, parsed via the appropriate
+/// language grammar, and the resulting [`tree_sitter::Tree`] is stored.
+/// Subsequent accesses return a clone of the cached tree.
+///
+/// When a tool modifies a file, call [`invalidate`](ParserCache::invalidate)
+/// to remove the stale entry — the next access will re-read and re-parse.
+#[derive(Debug, Default)]
+pub struct ParserCache {
+    /// Map from canonical file path → parsed tree.
+    cache: HashMap<PathBuf, tree_sitter::Tree>,
+}
+
+impl ParserCache {
+    /// Create an empty parser cache.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Return the parsed tree for `path`, re-parsing if not cached.
+    ///
+    /// The file extension determines which grammar to use. Unsupported
+    /// extensions return an [`io::Error`] with kind [`InvalidInput`].
+    ///
+    /// Returns a clone of the cached [`tree_sitter::Tree`] — the clone is
+    /// cheap (pointer copy). The returned `Tree` is owned and can be used
+    /// independently of the cache.
+    ///
+    /// # Errors
+    ///
+    /// * `io::ErrorKind::InvalidInput` — unsupported file extension.
+    /// * `io::ErrorKind::NotFound` or other I/O errors from reading the file.
+    /// * Parse failures return `io::ErrorKind::InvalidData`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the grammar `set_language` call fails after a valid
+    /// `language_for_path` lookup (should be unreachable).
+    pub fn parse_file(&mut self, path: &Path) -> io::Result<tree_sitter::Tree> {
+        if !self.cache.contains_key(path) {
+            let tree = self.do_parse(path)?;
+            self.cache.insert(path.to_path_buf(), tree);
+        }
+        // Tree::clone is a shallow copy (pointer clone to C-heap data).
+        Ok(self.cache[path].clone())
+    }
+
+    /// Invalidate the cached tree for `path`.
+    ///
+    /// The next call to [`parse_file`] will re-read and re-parse.
+    pub fn invalidate(&mut self, path: &Path) {
+        self.cache.remove(path);
+    }
+
+    /// Return the number of files with cached parse trees.
+    #[allow(dead_code)] // Only called from tests.
+    pub(crate) fn cache_size(&self) -> usize {
+        self.cache.len()
+    }
+
+    // -- internal helpers ----------------------------------------------------
+
+    /// Read, determine language, and parse a file.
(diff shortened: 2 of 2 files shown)


src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "non-UTF-8 path"))?;
+
+        let lang = language_for_path(path_str).ok_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("unsupported file extension for '{}'", path_str),
+            )
+        })?;
+
+        let content = std::fs::read_to_string(path)?;
```
decko: **Bug: `read_to_string` rejects non-UTF-8 files before tree-sitter even runs.**

Python files with a PEP 263 encoding declaration (e.g. `# -*- coding: latin-1 -*-`) or any non-ASCII bytes cause this to return `Err(InvalidData)` instead of a parse tree with error nodes.

tree-sitter's `parse()` accepts `impl AsRef<[u8]>`, so the fix is a one-liner:
```rust
let content = std::fs::read(path)?;
// parser.parse(&content, None) works unchanged — Vec<u8> satisfies AsRef<[u8]>
```

src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// * `io::ErrorKind::InvalidInput` — unsupported file extension.
+    /// * `io::ErrorKind::NotFound` or other I/O errors from reading the file.
+    /// * Parse failures return `io::ErrorKind::InvalidData`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the grammar `set_language` call fails after a valid
+    /// `language_for_path` lookup (should be unreachable).
+    pub fn parse_file(&mut self, path: &Path) -> io::Result<tree_sitter::Tree> {
+        if !self.cache.contains_key(path) {
```
decko: The struct field comment above says "canonical file path" but the key is inserted verbatim from the caller. Two paths referring to the same file (`src/main.rs` vs `./src/main.rs`) will create two cache entries and parse the file twice. See `src/tools/fs.rs:161` for how this codebase canonicalizes before cache operations:
```rust
let canonical = path.canonicalize()?;
if !self.cache.contains_key(&canonical) {
    let tree = self.do_parse(&canonical)?;
    self.cache.insert(canonical, tree);
}
```

src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+            let tree = self.do_parse(path)?;
+            self.cache.insert(path.to_path_buf(), tree);
+        }
+        // Tree::clone is a shallow copy (pointer clone to C-heap data).
+        Ok(self.cache[path].clone())
```
decko: Double lookup on the hot path: `contains_key` hashes once, then indexing hashes again. Use the entry API to hash once:
```rust
Ok(self.cache
    .entry(path.to_path_buf())
    .or_insert_with(|| self.do_parse(path).unwrap()) // or propagate error
    .clone())
```
(The error-propagating version needs a small helper since `or_insert_with` can't return `Result`, but the double-lookup is worth fixing.)

src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Invalidate the cached tree for `path`.
+    ///
+    /// The next call to [`parse_file`] will re-read and re-parse.
+    pub fn invalidate(&mut self, path: &Path) {
+        self.cache.remove(path);
+    }
+
+    /// Return the number of files with cached parse trees.
+    #[allow(dead_code)] // Only called from tests.
```
decko: `#[allow(dead_code)]` is a no-op on `pub(crate)` items — the compiler only warns `dead_code` for private items. Remove the attribute; if the intent was to keep this private, use `fn cache_size` (no visibility modifier).

src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+        let lang = language_for_path(path_str).ok_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("unsupported file extension for '{}'", path_str),
+            )
+        })?;
+
+        let content = std::fs::read_to_string(path)?;
+
+        let mut parser = tree_sitter::Parser::new();
```
decko: `Parser::new()` allocates a tree-sitter C parser on every cache miss and immediately drops it after the parse. For full-repo indexing (N distinct files = N alloc/free cycles), this adds up. Consider adding `parser: tree_sitter::Parser` as a field on `ParserCache` and resetting the language per call:
```rust
self.parser.set_language(&language_grammar(lang)).expect("...");
self.parser.parse(&content, None)
```

src/treesitter/parser.rs
```diff
@@ -0,0 +1,204 @@
...
+// Tests
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+
+    /// Write content to a temp file and return its path.
+    fn temp_file(name: &str, content: &str) -> PathBuf {
```
decko, line 141: All tests share a fixed directory (`carv-test-parser/`) with fixed filenames. `cargo test` runs tests within a binary concurrently by default — a future test that reuses e.g. `valid.rs` would race with `parse_rust_file`. Consider using the `tempfile` crate (`NamedTempFile`) or appending a unique suffix per test to lock this pattern against future additions.

src/treesitter/mod.rs
```diff
@@ -4,6 +4,8 @@
 //! via `tree_sitter_language::LanguageFn` constants. The module maps file
 //! extensions to [`Language`] variants and loads the corresponding grammar.
 
+pub mod parser;
```
decko: `pub mod parser` exposes `ParserCache` as part of the crate's public API, but it has no non-test callers and isn't wired into `ToolContext`. Any breaking change to `parse_file`'s signature will be a semver break once this ships. `pub(crate) mod parser` costs nothing and preserves the freedom to evolve the interface while it's still unwired.

decko (commented):
## Rust Specialist Review

Good foundation — the cache struct, invalidation API, and test coverage are well-shaped. Seven points below, one is a correctness bug, the rest are hardening and cleanup.

**Must-fix:**
- `read_to_string` breaks on non-UTF-8 Python files (see inline) — tree-sitter accepts `&[u8]` directly, use `fs::read` instead.

**Should-fix before merge:**
- Cache keys are not canonicalized despite the struct field comment saying "canonical file path". Use `path.canonicalize()?` before insert/lookup (see also how `fs.rs:161` handles this).
- `invalidate()` has zero non-test callers and `ParserCache` is never constructed outside `parser.rs` tests. The module docstring claims it fires via `AnchorState::notify_edit` — that wiring does not exist. This should either be wired in this PR or the doc should be corrected to match reality.
- `Parser::new()` is allocated and dropped on every cache miss. Move the parser into a `ParserCache` field and reuse it.

**Nits:**
- Double HashMap lookup (`contains_key` + index on line 62) → use the `entry` API.
- `#[allow(dead_code)]` on a `pub(crate)` item is a no-op — the compiler never warns on crate-visible items. Remove the attribute.
- `pub mod parser` in `mod.rs` exposes an unwired, still-evolving API publicly. `pub(crate)` is safer until there's a real consumer.

PR #99, opened 2026-06-02
feat(treesitter): SCM query files for Rust, Python, TypeScript (#32)

### LINKED ISSUE #32: G3-G4: Tree-sitter query files (rust.scm, python.scm, typescript.scm)
## Context
`src/treesitter/queries/`: S-expression query files for each language, embedded via `include_str!()`.

**Design ref:** `docs/designs/2026-04-25-carv-design.md`, Query files.

## Scope
- `rust.scm`: captures `@definition.function`, `@definition.method`, `@definition.struct`, `@definition.impl`, `@definition.trait`, `@name.definition.*`, `@name.reference`
- `python.scm`: captures `@definition.function`, `@definition.class`, `@name.definition.*`, `@name.reference`
- `typescript.scm`: captures `@definition.function`, `@definition.method`, `@definition.class`, `@definition.interface`, `@name.definition.*`, `@name.reference`

**Type:** SCM (S-expression, ~150 lines total)

**Test:** Unit: run each query against fixture file → captures function/class definitions

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
G3-G4: Tree-sitter S-expression query files with definition/reference captures. Embedded via include_str!().

## Files
- **rust.scm** — function, method, struct, enum, union, trait, impl, type alias, macro, module, constant/static definitions + call references
- **python.scm** — function, class, method, constant definitions + call references
- **typescript.scm** — function, method, class, interface, type alias, enum, arrow-function variable definitions + call references
- **queries/mod.rs** — query_for_language() dispatcher + 4 integration tests

## Capture names
- @definition.function
- @definition.method
- @definition.struct
- @definition.class
- @definition.interface
- @definition.trait
- @definition.impl
- @definition.macro
- @definition.module
- @definition.constant
- @name.definition.* (for name extraction in G5)
- @name.reference + @reference.call

## Verification
- [x] cargo build ✅
- [x] cargo test (248/248) ✅
- [x] cargo clippy ✅
- [x] cargo fmt ✅

Closes #32

### DIFF
--- a/src/treesitter/queries/mod.rs
+++ b/src/treesitter/queries/mod.rs
@@ -1,0 +1,160 @@
+//! Tree-sitter S-expression query files, embedded at compile time.
+//!
+//! Each language has a `.scm` file with `@definition.*` and `@name.*`
+//! captures used by the `get_skeleton`, `get_function`, and `replace_symbol`
+//! tools (G5, G6).
+
+use super::Language;
+
+// ---------------------------------------------------------------------------
+// Embedded queries
+// ---------------------------------------------------------------------------
+
+/// Return the tag query for `language` as a static string.
+///
+/// The query captures definitions (`@definition.function`,
+/// `@definition.class`, etc.) and references (`@name.reference`).
+#[allow(dead_code)] // Used by G5/G6 tree-sitter tools.
+pub fn query_for_language(lang: Language) -> &'static str {
+    match lang {
+        Language::Rust => include_str!("rust.scm"),
+        Language::Python => include_str!("python.scm"),
+        Language::TypeScript | Language::Tsx => include_str!("typescript.scm"),
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::treesitter::language_grammar;
+    use crate::treesitter::parser::ParserCache;
+    use std::io::Write;
+    use tree_sitter::{Query, StreamingIterator};
+
+    fn temp_file(name: &str, content: &str) -> std::path::PathBuf {
+        let dir = std::env::temp_dir().join("carv-test-queries");
+        std::fs::create_dir_all(&dir).unwrap();
+        let path = dir.join(name);
+        let mut f = std::fs::File::create(&path).unwrap();
+        f.write_all(content.as_bytes()).unwrap();
+        f.flush().unwrap();
+        path
+    }
+
+    /// Run a query and return a sorted list of capture names.
+    fn collect_captures(lang: Language, tree: &tree_sitter::Tree, content: &[u8]) -> Vec<String> {
+        let query_src = query_for_language(lang);
+        let language = language_grammar(lang);
+        let query = Query::new(&language, query_src).unwrap();
+        let mut cursor = tree_sitter::QueryCursor::new();
+        let mut captures: Vec<String> = Vec::new();
+        let mut matches = cursor.matches(&query, tree.root_node(), content);
+        while let Some(m) = matches.next() {
+            for cap in m.captures {
+                captures.push(query.capture_names()[cap.index as usize].to_string());
+            }
+        }
+        captures.sort();
+        captures
+    }
+
+    #[test]
+    fn rust_queries_capture_functions() {
+        let path = temp_file("query_rust.rs", "fn hello() {}\nfn world() {}\n");
+        let mut cache = ParserCache::new();
+        let tree = cache.parse_file(&path).unwrap();
+        let content = std::fs::read(&path).unwrap();
+
+        let captures = collect_captures(Language::Rust, &tree, &content);
+        let def_count = captures
+            .iter()
+            .filter(|c| c.as_str() == "definition.function")
+            .count();
+        assert_eq!(
+            def_count, 2,
+            "expected 2 function definitions, got: {:?}",
+            captures
+        );
+    }
+
+    #[test]
+    fn rust_queries_capture_structs() {
+        let path = temp_file("query_struct.rs", "struct Foo {}\nstruct Bar {}\n");
+        let mut cache = ParserCache::new();
+        let tree = cache.parse_file(&path).unwrap();
+        let content = std::fs::read(&path).unwrap();
+
+        let captures = collect_captures(Language::Rust, &tree, &content);
+        let def_count = captures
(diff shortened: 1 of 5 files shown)


src/treesitter/queries/rust.scm
```diff
@@ -0,0 +1,71 @@
...
+; Definitions
+; ---------------------------------------------------------------------------
+
+(function_item
+  name: (identifier) @name.definition.function) @definition.function
+
+; Method definitions (inside impl blocks).
+(declaration_list
+  (function_item
+    name: (identifier) @name.definition.method) @definition.method)
```
decko: **Bug: method double-capture.**

`function_item` at line 11 matches *any* `function_item` node — including those inside `impl` blocks. The `declaration_list` pattern below then also matches the same node. A method ends up captured as both `definition.function` and `definition.method`.

Fix: scope the free-function rule to the source file level, or add a `(#not-has-ancestor?)` predicate:
```scheme
; Free functions only
(source_file
  (function_item
    name: (identifier) @name.definition.function) @definition.function)
```
The test at `rust_queries_capture_all_definition_types` does not catch this because it does not assert on `definition.function` count — add an explicit check there.

decko (commented):
## Review Summary

**3 semantic bugs** (double-captures), 1 design issue in , and 1 test isolation risk. The dispatcher and embedding approach is solid — these are all in the  logic.

### Blocking
- Method double-capture in **Rust** and **Python**: methods match both `@definition.function` and `@definition.method` simultaneously
- Exported declaration double-capture in **TypeScript**: exported arrow functions / functions / classes match two patterns each

### Non-blocking
- `impl_item` emits two `@name.definition.impl` captures per match
- Shared temp dir with fixed filenames in tests (parallel-unsafe if filenames ever collide)
- No test coverage for exported TS declarations or method-only counts

Inline comments on the specific lines below.

src/treesitter/queries/rust.scm
```diff
@@ -0,0 +1,71 @@
...
+(trait_item
+  name: (type_identifier) @name.definition.trait) @definition.trait
+
+; Impl blocks.
+(impl_item
+  trait: (_)? @name.definition.impl
+  type: (type_identifier) @name.definition.impl) @definition.impl
+
+; Type aliases.
+(type_item
```
decko, line 43: **Design issue: two `@name.definition.impl` captures per match.**

For `impl Trait for Type`, both the `trait:` and `type:` fields emit `@name.definition.impl`. A downstream consumer iterating name captures will see two values per impl block.

`type` is always present and is the more stable identifier (the struct being implemented). Suggest keeping only that:
```scheme
(impl_item
  type: (type_identifier) @name.definition.impl) @definition.impl
```
If the trait name is also needed by G5, use a distinct capture name like `@name.definition.impl.trait`.

src/treesitter/queries/python.scm
```diff
@@ -0,0 +1,56 @@
...
+; Method definitions inside a class body.
+(class_definition
+  body: (block
+    (function_definition
+      name: (identifier) @name.definition.method) @definition.method))
+
+; Decorated functions / methods.
+(decorated_definition
+  definition: (function_definition
+    name: (identifier) @name.definition.function) @definition.function)
```
decko: **Bug: method double-capture.**

The top-level `(function_definition ...)` pattern (line 11) matches *any* function definition including those inside class bodies. The `(class_definition body: ...)` pattern below also matches the same node. `baz` in the test is captured as both `definition.function` and `definition.method`.

The test `python_queries_capture_functions_and_classes` passes only because it counts `fn_count == 2` (foo + baz), accepting the semantically incorrect double-capture of `baz`.

Fix — restrict the top-level rule to module-level functions using `(module ...)` as the ancestor, or add a negation predicate:
```scheme
; Module-level functions only
(module
  (function_definition
    name: (identifier) @name.definition.function) @definition.function)
```

src/treesitter/queries/typescript.scm
```diff
@@ -0,0 +1,76 @@
...
+    (interface_declaration
+      name: (type_identifier) @name.definition.interface) @definition.interface
+    (lexical_declaration
+      (variable_declarator
+        name: (identifier) @name.definition.function
+        value: (arrow_function)) @definition.function)
+  ])
+
+; Type aliases.
+(type_alias_declaration
```
decko, line 44: **Bug: exported declarations double-captured.**

The `(lexical_declaration ...)` pattern (line 33) matches all lexical declarations including exported ones. The `(export_statement declaration: (lexical_declaration ...))` block below also matches the same node. An `export const foo = () => {}` is captured as `definition.function` twice.

Same issue for exported `function_declaration` and `class_declaration` — each has both a standalone pattern and an entry inside `export_statement`.

No current test covers exported declarations so this is invisible in CI. Two options:
1. **Remove the standalone patterns** and always match through `export_statement` (loses non-exported declarations).
2. **Keep only standalone patterns** — tree-sitter walks the export wrapper's children anyway, so `(function_declaration ...)` already matches inside `(export_statement ...)`.

Option 2 is simpler and consistent with how the Rust/Python queries work.

src/treesitter/queries/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    use tree_sitter::{Query, StreamingIterator};
+
+    fn temp_file(name: &str, content: &str) -> std::path::PathBuf {
+        let dir = std::env::temp_dir().join("carv-test-queries");
+        std::fs::create_dir_all(&dir).unwrap();
+        let path = dir.join(name);
+        let mut f = std::fs::File::create(&path).unwrap();
+        f.write_all(content.as_bytes()).unwrap();
+        f.flush().unwrap();
+        path
```
decko, line 45: **Test isolation: shared temp dir with fixed filenames.**

Rust tests run in parallel by default. Current filenames (`query_rust.rs`, `query_all_rs.rs`, `query_py.py`, `query_ts.ts`) don't collide today, but any future test reusing a name will produce a race — one test overwrites the file mid-parse of another.

Prefer `tempfile::NamedTempFile` (avoids the manual flush/path dance too):
```rust
let mut f = tempfile::Builder::new().suffix(".rs").tempfile().unwrap();
f.write_all(content.as_bytes()).unwrap();
f.flush().unwrap();
let path = f.path().to_path_buf();
```
If adding `tempfile` as a dev-dependency is undesirable, at minimum append `std::thread::current().id()` to the filename.

PR #101, opened 2026-06-02
feat(treesitter): get_skeleton + get_function tools (#33)

### LINKED ISSUE #33: G5: get_skeleton + get_function tree-sitter tools
## Context
`src/tools/treesitter.rs`: Implement `get_skeleton` and `get_function` tools using tree-sitter queries.

## Scope
- `get_skeleton`: Parse file → run definition query → return signature lines with anchors
- `get_function`: Find named symbol via dot-path (`Class.method`) → walk parent nodes → extract body byte range → return body with anchors

**Type:** 🟠 Dense (~150 lines)

**Test:** Unit: skeleton on fixture .rs → returns function signatures; get_function("main") → returns body

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
Implements G5 tree-sitter structural tools: `get_skeleton` and `get_function`.

**get_skeleton:** Parses a source file, runs the language-appropriate definition query, and returns anchored signature lines per definition (functions, structs, classes, etc.).

**get_function:** Finds a symbol by name or dot-path (`Class.method`), walks the AST, extracts its body byte range, and returns the body with stable anchors.

## Changes
- `src/tools/treesitter.rs` (new) — GetSkeletonTool + GetFunctionTool (904 lines, ~400 lines tests)
- `src/treesitter/queries/` (new) — rust.scm, python.scm, typescript.scm, mod.rs
- `src/tools/traits.rs` — Added `parser_cache` to ToolContext (user-approved)
- `src/treesitter/parser.rs` — Added `Default` impl for ParserCache
- `src/tools/fs.rs`, `src/tools/edit.rs` — Wired ParserCache invalidation
- `src/tools/mod.rs` — Registered new tools in `default_tools()`
- All test contexts updated across `tools/fs.rs`, `exec.rs`, `search.rs`, `test_utils.rs`, `traits.rs`

## Review fixes applied
- ✅ **Anchor occurrence indices (Invariant #3):** Both tools now use `AnchorState::get_anchors()` for occurrence-indexed anchors that match `read_file`/`edit_file` output
- ✅ **Lock poisoning (Invariant #2):** Replaced `.expect()` with `match` + `ToolResult::error`
- ✅ **Cache invalidation:** Wired `ParserCache::invalidate()` into `write_file` and `edit_file`

## Verification
- 263 tests (all pass), 0 clippy warnings, fmt clean
- 15 unit tests for treesitter tools (Rust/Python/TS)
- 4 query integration tests

Closes #33

### DIFF
--- a/src/tools/fs.rs
+++ b/src/tools/fs.rs
@@ -162,6 +162,11 @@ impl Tool for WriteFileTool {
                     let mut anchor_state =
                         ctx.anchor_state.lock().expect("anchor state lock poisoned");
                     anchor_state.notify_edit(&canonical);
+                    // Also invalidate the tree-sitter parser cache so
+                    // get_skeleton / get_function return fresh ASTs.
+                    if let Ok(mut pc) = ctx.parser_cache.lock() {
+                        pc.invalidate(&canonical);
+                    }
                     Ok(ToolResult::ok(format!(
                         "Wrote {byte_count} bytes to {}",
                         canonical.display()
--- a/src/tools/fs.rs
+++ b/src/tools/fs.rs
@@ -515,6 +515,7 @@ mod tests {
         let ctx = ToolContext {
             workspace_root: dir,
             anchor_state: anchor_state.clone(),
+            parser_cache: Arc::new(Mutex::new(crate::treesitter::parser::ParserCache::new())),
         };
 
         // Write via the tool — this should invalidate the cache.
--- a/src/tools/traits.rs
+++ b/src/tools/traits.rs
@@ -79,12 +79,18 @@ impl ToolResult {
 /// Context shared across all tool invocations.
 ///
 /// Carries project state accessible to every tool's `execute` method.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ToolContext {
     /// Absolute path to the project root directory.
     pub workspace_root: PathBuf,
     /// Shared anchor state for hash-anchored line referencing.
     pub anchor_state: Arc<Mutex<crate::hashing::state::AnchorState>>,
+    /// Shared tree-sitter parser cache for per-file AST caching.
+    ///
+    /// Added for G5 (get_skeleton / get_function).  Previous tools
+    /// (read_file, edit_file, etc.) do not use tree-sitter so this
+    /// field is unused by them.
+    pub parser_cache: Arc<Mutex<crate::treesitter::parser::ParserCache>>,
 }
 
 // ---------------------------------------------------------------------------
@@ -167,6 +173,7 @@ mod tests {
         ToolContext {
             workspace_root: PathBuf::from("/tmp/test"),
             anchor_state: Arc::new(Mutex::new(crate::hashing::state::AnchorState::new())),
+            parser_cache: Arc::new(Mutex::new(crate::treesitter::parser::ParserCache::new())),
         }
     }
 
--- a/src/tools/traits.rs
+++ b/src/tools/traits.rs
@@ -79,7 +79,7 @@ impl ToolResult {
 /// Context shared across all tool invocations.
 ///
 /// Carries project state accessible to every tool's `execute` method.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ToolContext {
     /// Absolute path to the project root directory.
     pub workspace_root: PathBuf,
--- a/src/tools/treesitter.rs
+++ b/src/tools/treesitter.rs
@@ -168,22 +168,26 @@ impl Tool for GetSkeletonTool {
                 }
             };
 
-            // Parse the file.
-            let mut parser_cache = match ctx.parser_cache.lock() {
-                Ok(g) => g,
-                Err(_) => {
-                    return Ok(ToolResult::error(
-                        "get_skeleton failed: parser cache lock poisoned",
-                    ));
-                }
-            };
-            let tree = match parser_cache.parse_file(&resolved) {
-                Ok(t) => t,
-                Err(e) => {
-                    return Ok(ToolResult::error(format!(
-                        "get_skeleton failed: cannot parse '{}': {e}",
-                        resolved.display()
-                    )));
+            // Parse the file. Lock parser cache only for the parse; release
+            // immediately afterward so we don't hold the lock during anchor
+            // state operations and file I/O.
+            let tree = {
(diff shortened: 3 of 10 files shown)


src/tools/treesitter.rs
```diff
@@ -0,0 +1,964 @@
...
+/// Find a body node inside a definition node.
+///
+/// Walks named children looking for a body-type node (`block`,
+/// `statement_block`, etc.) and returns it.
+fn find_body_node(node: tree_sitter::Node, lang: Language) -> Option<tree_sitter::Node> {
+    let kinds = body_kinds(lang);
+    for i in 0..node.child_count() {
+        if let Some(child) = node.child(i as u32) {
+            if child.is_named() && kinds.contains(&child.kind()) {
+                return Some(child);
```
decko, line 335: **Lock held across I/O.** The `parser_cache` `MutexGuard` is acquired at line 312 and not dropped until the end of the `async move` block — it stays held through `read_file_bytes` here and all subsequent work including the `anchor_state` lock. If `parse_file` returns an owned `tree_sitter::Tree`, add an explicit `drop(parser_cache)` after obtaining the tree so the lock is released before doing any I/O. Same issue applies to `GetFunctionTool::execute`.

src/tools/treesitter.rs
```diff
@@ -0,0 +1,964 @@
...
+                        "get_function failed: dot-path must have exactly 2 parts (e.g. 'Class.method'), got '{}'",
+                        symbol
+                    )));
+                }
+                find_dotpath_symbol(tree.root_node(), &parts, &content, lang)
+            } else {
+                find_definition_by_name(tree.root_node(), symbol, &content, lang)
+            };
+
+            let def_node = match def_node {
```
decko, line 538: **Unguarded index access.** `parts[1]` panics if `parts` has fewer than 2 elements. The `execute` call site guards for `parts.len() != 2`, but this function takes `&[&str]` and could be called with any slice. Consider changing the signature to `(node, outer: &str, inner: &str, content, lang)` to make the two-part invariant explicit at the type level, or add a defensive `if parts.len() < 2 { return None; }` at the top.

src/tools/treesitter.rs
```diff
@@ -0,0 +1,964 @@
...
+        serde_json::json!({
+            "type": "object",
+            "properties": {
+                "path": {
+                    "type": "string",
+                    "description": "Path to the source file, relative to the project root or absolute"
+                },
+                "symbol": {
+                    "type": "string",
+                    "description": "Name of the function/method to extract. Use dot-notation for methods (e.g. 'MyStruct.my_method')"
```
decko, line 456: **Non-idiomatic child iteration.** This `for i in 0..node.child_count()` + `node.child(i as u32)` pattern appears in all four tree-walking helpers. `tree_sitter::Node` exposes `named_children(&mut cursor)` which avoids the index cast and is the idiomatic approach. A single `TreeCursor` can be reused across calls.

src/tools/traits.rs
```diff
@@ -79,12 +79,18 @@ impl ToolResult {
 /// Context shared across all tool invocations.
 ///
 /// Carries project state accessible to every tool's `execute` method.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
```
decko: **`Clone` unnecessarily removed.** `Arc<Mutex<ParserCache>>` implements `Clone` regardless of whether `ParserCache: Clone`, so the derive could be kept. Dropping it is a silent breaking change for any future caller that clones `ToolContext`. If removal is intentional, add a comment explaining why.

src/tools/fs.rs
```diff
@@ -162,6 +162,11 @@ impl Tool for WriteFileTool {
                     let mut anchor_state =
                         ctx.anchor_state.lock().expect("anchor state lock poisoned");
                     anchor_state.notify_edit(&canonical);
+                    // Also invalidate the tree-sitter parser cache so
+                    // get_skeleton / get_function return fresh ASTs.
+                    if let Ok(mut pc) = ctx.parser_cache.lock() {
```
decko, line 167: **Inconsistent lock-poison handling.** `if let Ok` silently swallows a poisoned `parser_cache` lock, leaving the cache stale with no signal to the caller. Compare: `get_skeleton`/`get_function` return `ToolResult::error` on lock poison, and `anchor_state.lock()` two lines above uses `.expect()`. At minimum log a warning on `Err(_)` so cache staleness is observable.

src/treesitter/parser.rs
```diff
@@ -50,7 +50,18 @@ impl ParserCache {
...
+
+// --- Default impl ---------------------------------------------------------
 
+impl Default for ParserCache {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[allow(dead_code)] // Methods used only from tests; external callers in G5-G6.
```
decko, line 63: **`#[allow(dead_code)]` is likely stale.** The comment says "external callers in G5-G6", but G5 is this PR — `parse_file` and `invalidate` are now called in production code. Either remove the attribute or scope it to only the methods that remain G6-only. A blanket suppression on the entire `impl` block will hide future unused-method warnings.

decko (commented):
## Overview

Adds `get_skeleton` (file definition outline) and `get_function` (symbol body extraction) using tree-sitter, with a shared `ParserCache` wired into `ToolContext`. Cache invalidation is hooked into `write_file`/`edit_file`. Three `.scm` query files for Rust, Python, and TypeScript. Well-structured with good test coverage overall.

---

## Missing Test Cases

- **Duplicate symbol names** — two `fn foo()` in the same file: which does `get_function` return? First-match semantics should be explicitly verified.
- **Decorated Python functions** — the `.scm` has a `decorated_definition` pattern but no test exercises it.
- **Cache invalidation roundtrip** — write a file, call `get_skeleton`, edit it, call `get_skeleton` again; verify the result reflects the edit. There is a cache invalidation unit test in `fs.rs` but no end-to-end tool-level test.
- **`get_skeleton` with nested `mod`** — does the Rust query surface definitions inside inline modules?

---

## Summary

Main actionable items: over-wide parser cache lock scope and the `Clone` regression on `ToolContext`. The `parts[1]` access is technically safe today due to the caller guard but is a landmine for future refactors. Everything else is low-priority polish.

PR #102, opened 2026-06-02
feat(treesitter): replace_symbol tool (#34)

### LINKED ISSUE #34: G6: replace_symbol tree-sitter tool
## Context
`src/tools/treesitter.rs`: Implement `replace_symbol` — replace a function/class by AST node with byte-range splicing.

## Scope
- Find symbold AST node → byte range (extending to include decorators/export wrappers)
- Splice in new code at byte range
- Multiple replacements applied bottom-to-top (avoid offset corruption)
- Invalidate anchors + tree-sitter cache after edit

**Type:** 🟠 Dense (~120 lines)

**Test:** Unit: replace_symbol("old_fn", "new_fn") → file content updated, anchors invalidated

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

decko:
## Summary
Implements G6: `replace_symbol` tool — AST-aware function/class replacement with byte-range splicing.

**Features:**
- Finds symbol by name or dot-path (`Class.method`) via existing AST traversal
- Extends byte range upward through wrapper nodes (Python decorators, TS export statements)
- Splices new code at the computed byte range
- Multi-file batching via `files` array
- Bottom-to-top ordering within each file to preserve byte offsets
- Invalidates anchor state + parser cache after each write

**Wrapper extension:**
- Python: `decorated_definition`
- TypeScript: `export_statement`, `export_default_declaration`

## Tests (8 new)
- Replace function (Rust)
- Method via dot-path (Rust)
- Cache invalidation after replace
- Missing symbol error
- Empty files array
- Python decorated function (wrapper removal)
- TypeScript exported function (wrapper removal)
- Multi-file same path (bottom-to-top ordering)

## Verification
- 271 tests (all pass), 0 clippy warnings, fmt clean

Closes #34

### DIFF
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -65,5 +65,6 @@ pub fn default_tools() -> Vec<Box<dyn Tool>> {
         Box::new(search::SearchFilesTool),
         Box::new(treesitter::GetSkeletonTool),
         Box::new(treesitter::GetFunctionTool),
+        Box::new(treesitter::ReplaceSymbolTool),
     ]
 }
--- a/src/tools/treesitter.rs
+++ b/src/tools/treesitter.rs
@@ -608,6 +608,303 @@ impl Tool for GetFunctionTool {
 }
 
 // ---------------------------------------------------------------------------
+// replace_symbol
+// ---------------------------------------------------------------------------
+
+/// Parent node kinds that wrap a definition and should be included when
+/// computing the byte range for replacement.
+///
+/// Example: a Python function decorated with `@decorator` is wrapped in a
+/// `decorated_definition` node. Replacing just the `function_definition`
+/// child would leave the orphaned decorator — we extend the range upward.
+const WRAPPER_KINDS: &[&str] = &[
+    // Python: @decorator wrapping a function or class
+    "decorated_definition",
+    // TypeScript: `export function ...`, `export default class ...`
+    "export_statement",
+    "export_default_declaration",
+];
+
+/// Extend the byte range of a definition node upward to include wrapper
+/// nodes (decorators, export statements) that should be part of the
+/// replacement.
+///
+/// Walks up from `node` through the parent chain; for each parent whose
+/// `kind()` is in [`WRAPPER_KINDS`], the range is widened to include it.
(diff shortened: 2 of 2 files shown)


decko:
## Rust Specialist Review

Overall solid implementation — the bottom-to-top byte offset strategy is correct, overlap detection works, and lock scopes are tight. A few issues worth addressing:

### Critical

**Phase 2 failures silently return `ToolResult::ok`**

When individual edits fail in phase 2 (file read error, byte range OOB, write error), errors are collected into `results` and the function returns `ToolResult::ok(results.join("\n"))`. A caller receiving `"replace_symbol failed for 'foo' in main.rs: ..."` inside an `ok` result won't detect the failure without string-parsing. Should track whether any edit failed and return `ToolResult::error` accordingly.

### Minor

**`new()` is a reserved keyword in `replace_symbol_typescript_exported_function`**

The replacement code is `export function new() { return 1; }` — `new` is reserved in TS/JS, so the test produces invalid TypeScript. The assertion only checks for the string `"new()"`, so it passes, but the output is semantically wrong. Use a non-reserved name.

**`Err(_e)` bindings go unused**

Both cache invalidation blocks use `Err(_e)` where `_e` is never referenced. Should be `Err(_)`.

### Design

**Inconsistent phase 1 vs phase 2 error semantics** — Phase 1 returns early on first error (no writes occur). Phase 2 collects errors and continues. The asymmetry is defensible but undocumented; callers should know partial-failure leaves some files modified.

**`ResolvedEdit` struct defined inside `execute`** — valid Rust, but a local struct inside an async closure limits reuse. Module scope with `pub(super)` would be cleaner.

**O(n²) overlap check** — fine for typical workloads, a sort + linear scan would be more correct at scale.

### What's solid

- Bottom-to-top ordering is correct: descending byte offset sort ensures prior edits don't shift offsets for subsequent ones ✓  
- Overlap detection (`a.end > b.start`) correctly catches nested ranges, confirmed by the `mod foo` / `bar` test ✓  
- Parser cache lock dropped immediately after `parse_file` — no lock held across `await` ✓  
- Workspace path check happens before any parsing ✓  
- Test coverage is good (10 tests, including path traversal and overlap rejection) ✓

Issue #1, opened 2026-04-27
Create AGENTS.md project guide for AI coding agents


decko:
## Context

This project needs an `AGENTS.md` file at the repository root to provide tool-agnostic guidance for AI coding agents (Claude Code, Cursor, Copilot, etc.) working on the codebase. It complements the existing OpenCode-specific agent config in `.opencode/agents/`.

## What to include

- Build/test/lint commands
- Architecture overview (module tree)
- Critical invariants (no async-trait, hash-anchored editing, etc.)
- Mandatory git workflow (worktrees, no commits to main)
- Definition of Done checklist with reviewer gate
- Ticket assignment rule
- Complete workflow summary
- Error handling philosophy
- Code style
- Testing strategy
- Dependency notes

## Acceptance Criteria

- [ ] `AGENTS.md` exists at repo root
- [ ] Content covers all items in the design doc summary
- [ ] Git workflow rules are present (no commits to main, worktree-only)
- [ ] DoD reviewer gate is documented

## DoD Checklist
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt
- [ ] DoD review passed

Issue #36, opened 2026-04-27
H2: Token budget tracking + context window truncation at 80%


decko:
## Context
`src/agent/budget.rs`: Track cumulative token usage, truncate old results when approaching context window limits.

**Design ref:** `docs/designs/2026-04-25-carv-design.md`, Token Budget Management.

## Scope
- Track cumulative input/output/cache tokens across turns
- Estimate message payload size before each LLM call
- At 80% window: truncate old tool results → keep last N in full, summarize older to one line
- If still too large: drop oldest conversation turns entirely

**Type:** 🟠 Dense (~100 lines)

**Test:** Unit: simulate 20 turns → budget exceeds 80% → old results truncated

## DoD
- [ ] cargo build
- [ ] cargo test
- [ ] cargo clippy
- [ ] cargo fmt

PR #105, opened 2026-06-14
feat(agent): token budget tracking (H2, #36)

decko:
## Summary
Implement `TokenBudget` in `src/agent/budget.rs` — tracks cumulative token usage and provides context window threshold checks.

## Changes
- **`src/agent/budget.rs`** (new, 186 lines): Token budget tracker with:
  - `TokenBudget::new(context_window)` — 80% threshold by default
  - `record_usage(&LlmUsage)` — accumulates input, output, and cache read tokens
  - `estimate(json_str)` — heuristic: 4 chars ≈ 1 token
  - `is_over_threshold(estimated)` — checks against 80% window
  - Accessors: `context_window()`, `threshold()`, `total_input()`, `total_output()`
- **`src/agent/mod.rs`**: Added `pub mod budget;`

## DoD
- [x] cargo build ✅
- [x] cargo test (294 tests, 9 new) ✅
- [x] cargo clippy -- -D warnings ✅
- [x] cargo fmt -- --check ✅
- [x] No new dependencies
- [x] No API changes
- [x] No security boundary changes
- [x] 9 unit tests: defaults, accumulation, cache tokens, threshold boundaries, window sizes

Part of #36

### DIFF
--- a/src/agent/budget.rs
+++ b/src/agent/budget.rs
@@ -1,0 +1,185 @@
+//! Token budget tracking and context window truncation.
+//!
+//! Tracks cumulative token usage across agent turns and truncates
+//! conversation history when approaching the model's context window
+//! (80% threshold by default).
+
+use crate::llm::types::LlmUsage;
+
+/// Token budget tracker for context window management.
+///
+/// Before each LLM call, the agent estimates the message payload size
+/// and truncates old tool results if the threshold is exceeded. If
+/// truncating tool results isn't enough, the oldest conversation
+/// turns are dropped entirely.
+pub struct TokenBudget {
+    context_window: usize,
+    threshold: f64,
+    total_input: u64,
+    total_output: u64,
+}
+
+impl TokenBudget {
+    /// Create a new budget with the given context window size.
+    ///
+    /// Defaults: 80% threshold.
+    pub fn new(context_window: usize) -> Self {
+        Self {
+            context_window,
+            threshold: 0.8,
+            total_input: 0,
+            total_output: 0,
+        }
+    }
+
+    /// Record token usage from a completed turn.
+    pub fn record_usage(&mut self, usage: &LlmUsage) {
+        self.total_input += usage.input_tokens as u64;
+        self.total_output += usage.output_tokens as u64;
+        if let Some(cache) = usage.cache_read_tokens {
+            self.total_input += cache as u64;
+        }
+    }
+
+    /// Estimated token count for a messages array.
+    ///
+    /// Uses a simple heuristic: ~4 characters per token for English text.
+    /// Not exact, but sufficient for the 80% threshold check.
+    pub fn estimate(&self, messages_json: &str) -> usize {
+        messages_json.len() / 4
+    }
+
+    /// Whether the estimated payload exceeds the threshold.
+    pub fn is_over_threshold(&self, estimated: usize) -> bool {
+        let limit = (self.context_window as f64 * self.threshold) as usize;
+        estimated >= limit
+    }
+
+    /// Context window size.
+    pub fn context_window(&self) -> usize {
+        self.context_window
+    }
+
+    /// Current threshold fraction.
+    pub fn threshold(&self) -> f64 {
+        self.threshold
+    }
+
+    /// Total input tokens across all turns.
+    pub fn total_input(&self) -> u64 {
+        self.total_input
+    }
+
+    /// Total output tokens across all turns.
+    pub fn total_output(&self) -> u64 {
+        self.total_output
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn make_usage(input: u32, output: u32) -> LlmUsage {
+        LlmUsage {
+            input_tokens: input,
+            output_tokens: output,
+            cache_read_tokens: None,
+            cache_creation_tokens: None,
+        }
+    }
+
+    #[test]
+    fn test_new_budget_defaults() {
+        let budget = TokenBudget::new(200_000);
+        assert_eq!(budget.context_window(), 200_000);
+        assert!((budget.threshold() - 0.8).abs() < f64::EPSILON);
+        assert_eq!(budget.total_input(), 0);
+        assert_eq!(budget.total_output(), 0);
+    }
+
+    #[test]
+    fn test_record_usage_accumulates() {
+        let mut budget = TokenBudget::new(100_000);
+        budget.record_usage(&make_usage(1000, 500));
+        budget.record_usage(&make_usage(2000, 800));
+        assert_eq!(budget.total_input(), 3000);
+        assert_eq!(budget.total_output(), 1300);
+    }
+
+    #[test]
+    fn test_record_usage_with_cache_read() {
+        let mut budget = TokenBudget::new(100_000);
+        let usage = LlmUsage {
+            input_tokens: 1000,
+            output_tokens: 500,
+            cache_read_tokens: Some(300),
+            cache_creation_tokens: Some(100),
+        };
+        budget.record_usage(&usage);
+        // cache_read counts toward input
+        assert_eq!(budget.total_input(), 1300);
+        assert_eq!(budget.total_output(), 500);
+    }
+
+    #[test]
+    fn test_estimate_approximates_tokens() {
+        let budget = TokenBudget::new(100_000);
+        let json = r#"{"messages":[{"role":"user","content":"hello world"}]}"#;
+        let tokens = budget.estimate(json);
+        // ~74 chars / 4 ≈ 18 tokens
+        assert!(tokens > 10, "estimated {tokens} should be > 10");
+        assert!(tokens < 30, "estimated {tokens} should be < 30");
+    }
+
+    #[test]
+    fn test_is_over_threshold_below() {
+        let budget = TokenBudget::new(100_000);
+        // 100k * 0.8 = 80k.  10k is well below.
+        assert!(!budget.is_over_threshold(10_000));
+    }
+
+    #[test]
+    fn test_is_over_threshold_at_boundary() {
+        let budget = TokenBudget::new(100_000);
+        // 100k * 0.8 = 80k.  79_999 is below, 80_000 is at threshold.
+        assert!(!budget.is_over_threshold(79_999));
+        assert!(budget.is_over_threshold(80_000));
+    }
+
+    #[test]
+    fn test_is_over_threshold_above() {
+        let budget = TokenBudget::new(100_000);
+        assert!(budget.is_over_threshold(90_000));
+    }
+
+    #[test]
+    fn test_total_tokens_includes_cache() {
+        let mut budget = TokenBudget::new(200_000);
+        // 3 turns with cache hits
+        for _ in 0..3 {
+            let usage = LlmUsage {
+                input_tokens: 2000,
+                output_tokens: 1000,
+                cache_read_tokens: Some(1500),
+                cache_creation_tokens: None,
+            };
+            budget.record_usage(&usage);
+        }
+        // input = 3 * (2000 + 1500) = 10500
+        assert_eq!(budget.total_input(), 10500);
+        assert_eq!(budget.total_output(), 3000);
+    }
+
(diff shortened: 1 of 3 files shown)


decko:
Review findings addressed in [911d559](https://github.com/decko/carv/commit/911d559):

1. **🔴 Cache read double-counting** → Fixed. `record_usage` no longer adds `cache_read_tokens` separately — `input_tokens` already includes them per Anthropic wire format. Added `test_record_usage_cache_read_not_double_counted`.
2. **🔴 `estimate` API doesn't include tools** → Renamed to `estimate_payload()`, now accepts `&[Message]` + `&[ToolDef]`, serializes internally. Added `test_estimate_payload_includes_tools` verifying tools increase estimate.
3. **🟡 Module doc over-promises truncation** → Reworded: "Truncation logic lives in the loop module (H3) — the budget only reports whether the threshold is exceeded."
4. **🟡 Cache creation tokens ignored** → Added `total_cache_creation` field + `total_cache_creation()` accessor for billing summary.

294 tests pass, all 4 cargo checks green.

src/agent/budget.rs
```diff
@@ -0,0 +1,216 @@
...
+        if let Ok(json) = serde_json::to_string(tools) {
+            len += json.len();
+        }
+        len / 4
+    }
+
+    /// Whether the estimated payload exceeds the threshold.
+    pub fn is_over_threshold(&self, estimated: usize) -> bool {
+        let limit = (self.context_window as f64 * self.threshold) as usize;
+        estimated >= limit
```
decko: **Bug: `is_over_threshold` ignores cumulative context usage**

This method compares only the *current estimated payload* against the 80% limit, but never looks at `total_input` or `total_output` accumulated via `record_usage`. Those fields are collected but never consulted here — they are effectively dead state.

Failure scenario: after 20 short turns, `total_input` grows to 95% of `context_window`. The next call with a small payload returns `false`, the agent sends another turn, and the LLM API returns a context-length error.

```rust
pub fn is_over_threshold(&self, estimated: usize) -> bool {
    let limit = (self.context_window as f64 * self.threshold) as usize;
    let cumulative = (self.total_input + self.total_output) as usize;
    estimated >= limit || cumulative >= limit
}
```

src/agent/budget.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    /// Estimate the token count of a message payload including tools.
+    ///
+    /// Uses a simple heuristic: ~4 characters per token for English text.
+    /// Serializes messages and tools to JSON internally.
+    pub fn estimate_payload(&self, messages: &[Message], tools: &[ToolDef]) -> usize {
+        let mut len = 0;
+        if let Ok(json) = serde_json::to_string(messages) {
+            len += json.len();
+        }
```
decko: **Bug: serialization failure silently yields 0 for that component**

If `serde_json::to_string(messages)` returns `Err`, the branch is skipped, `len` stays 0 for messages, and `estimate_payload` returns a gross underestimate. `is_over_threshold` then returns `false`, the agent sends an oversized payload, and the API call fails.

In practice `Message` and `ToolDef` should always serialize cleanly, but the silent discard makes failures invisible. Prefer an `unwrap_or` with a conservative fallback or surface the error to the caller:

```rust
pub fn estimate_payload(&self, messages: &[Message], tools: &[ToolDef]) -> usize {
    let msg_bytes = serde_json::to_string(messages).map(|s| s.len()).unwrap_or(usize::MAX / 2);
    let tool_bytes = serde_json::to_string(tools).map(|s| s.len()).unwrap_or(usize::MAX / 2);
    (msg_bytes + tool_bytes) / 4
}
```

src/agent/budget.rs
```diff
@@ -0,0 +1,216 @@
...
+//! threshold checks for the core loop. Truncation logic lives in
+//! the loop module (H3) — the budget only reports whether the
+//! threshold is exceeded.
+
+use crate::llm::types::{LlmUsage, Message, ToolDef};
+
+/// Token budget tracker for context window management.
+///
+/// Before each LLM call, the agent estimates the message payload size
+/// (system prompt + conversation history + tools) and checks whether
```
decko: **Doc/correctness mismatch: system prompt is absent from `estimate_payload`**

The struct-level doc claims the estimate covers "system prompt + conversation history + tools", but `estimate_payload` only receives `messages` and `tools` — the system prompt is never passed in. If the system prompt is large (e.g. the one built by `build_system_prompt` in `context.rs`), the real token count at the LLM boundary consistently exceeds the estimate, causing `is_over_threshold` to underfire.

Either update the signature to accept a `system_prompt: &str` and include its length, or correct the doc to reflect what is actually measured.

src/agent/budget.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+        if let Ok(json) = serde_json::to_string(tools) {
+            len += json.len();
+        }
+        len / 4
+    }
+
+    /// Whether the estimated payload exceeds the threshold.
+    pub fn is_over_threshold(&self, estimated: usize) -> bool {
+        let limit = (self.context_window as f64 * self.threshold) as usize;
```
decko: **Edge case: `context_window = 0` causes `is_over_threshold` to always return `true`**

`(0_usize as f64 * 0.8) as usize` evaluates to `0`. Since `estimated` is `usize`, `estimated >= 0` is always `true`. Every call aborts the agent immediately before sending a single token.

Same behavior for `context_window = 1` (`limit = 0` after truncation).

```rust
// Guard at construction time:
pub fn new(context_window: usize) -> Self {
    assert!(context_window > 0, "context_window must be > 0");
    ...
}
```

decko (commented):
## Rust Specialist Review — `feat(agent): token budget tracking (H2, #36)`

4 inline findings were posted. Summary ranked by severity:

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **High** | `is_over_threshold` L65–67 | Ignores `total_input`/`total_output` — cumulative context pressure is invisible; `record_usage` data is collected but never used in the threshold check |
| 2 | **Medium** | `estimate_payload` L57 | Silent `serde_json` error swallow returns 0 for the failed component, causing systematic underestimation with no observable signal |
| 3 | **Medium** | Struct doc L13 | Doc claims estimate includes system prompt; `estimate_payload` signature takes no system prompt — real payload is always higher than estimated |
| 4 | **Low** | `is_over_threshold` L66 | `context_window = 0` or `= 1` → `limit = 0` → always returns `true`, aborting the agent before any turn |

**Overall**: The most structural issue is #1 — `record_usage` and `is_over_threshold` are two parallel mechanisms that never communicate. The budget tracks cumulative usage faithfully but the threshold check is blind to it, defeating the purpose of accumulation. Issues #2 and #3 both push estimates downward, compounding the risk of the budget underfiring.

decko:
All 4 findings addressed in 85b0eda:

1. is_over_threshold ignores cumulative usage -> Now checks both estimated payload AND cumulative total_input+total_output against the 80% limit. Added 2 tests: cumulative trigger, cumulative not-yet-reached.

2. Silent serde_json error swallow -> Changed from if-let-Ok to map+unwrap_or(usize::MAX/2): serialization failures now produce a conservative estimate instead of silently undercounting.

3. Doc/code mismatch: system prompt absent -> Added system_prompt parameter to estimate_payload. Fixed struct doc. Added test_estimate_payload_includes_system_prompt.

4. context_window=0 causes always-true -> Added assert!(>0) in new(). Added #[should_panic] test.

298 tests pass, all 4 cargo checks green.

src/agent/budget.rs
```diff
@@ -0,0 +1,270 @@
...
+        tools: &[ToolDef],
+    ) -> usize {
+        let sys_len = system_prompt.len();
+        let msg_len = serde_json::to_string(messages)
+            .map(|s| s.len())
+            .unwrap_or(usize::MAX / 2);
+        let tool_len = serde_json::to_string(tools)
+            .map(|s| s.len())
+            .unwrap_or(usize::MAX / 2);
+        (sys_len + msg_len + tool_len) / 4
```
decko: **Bug: double serialization failure causes integer overflow before `/4`**

The conservative fallback (`usize::MAX / 2`) only works safely when at most *one* serialization fails. If both fail:

```
msg_len  = usize::MAX / 2  (~9.2e18 on 64-bit)
tool_len = usize::MAX / 2
sys_len + msg_len + tool_len  →  overflows usize
```

- **Debug mode**: panics on the addition
- **Release mode**: wraps silently to `sys_len - 1`, then `/4` → tiny number → `is_over_threshold` returns `false`

The intent (overestimate on failure) is inverted in release builds.

```rust
// Fix: saturating addition prevents overflow in both modes
(sys_len.saturating_add(msg_len).saturating_add(tool_len)) / 4
```

src/agent/budget.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /// Whether the estimated payload or cumulative usage exceeds the
+    /// threshold. Considers both the current payload estimate and the
+    /// running total from `record_usage` calls.
+    pub fn is_over_threshold(&self, estimated: usize) -> bool {
+        let limit = (self.context_window as f64 * self.threshold) as usize;
+        let cumulative = (self.total_input + self.total_output) as usize;
```
decko: **Bug: `u64 → usize` narrowing cast truncates silently on 32-bit targets**

`total_input` and `total_output` are `u64`. On `wasm32` or any 32-bit target `usize` is 32 bits. Once cumulative usage exceeds ~4 billion tokens (`u32::MAX`), the cast wraps silently and `cumulative` appears near-zero — the threshold check never fires.

```rust
// Fix: use saturating_add and an explicit u64 comparison
let limit_u64 = (self.context_window as f64 * self.threshold) as u64;
let cumulative = self.total_input.saturating_add(self.total_output);
(estimated as u64) >= limit_u64 || cumulative >= limit_u64
```

src/agent/budget.rs
```diff
@@ -0,0 +1,270 @@
...
+            .map(|s| s.len())
+            .unwrap_or(usize::MAX / 2);
+        (sys_len + msg_len + tool_len) / 4
+    }
+
+    /// Whether the estimated payload or cumulative usage exceeds the
+    /// threshold. Considers both the current payload estimate and the
+    /// running total from `record_usage` calls.
+    pub fn is_over_threshold(&self, estimated: usize) -> bool {
+        let limit = (self.context_window as f64 * self.threshold) as usize;
```
decko: **Edge case: `context_window = 1` still makes `limit = 0` → always true**

The `assert!(context_window > 0)` in `new` guards against zero but not against 1:

```rust
(1_usize as f64 * 0.8) as usize  // = 0 (truncation)
```

`limit = 0`, so `estimated >= 0` is always `true` for `usize`. The agent aborts immediately on every turn.

The assert should reject degenerate values that produce a zero limit after truncation:

```rust
assert!(context_window >= 100, "context_window too small to be meaningful");
// or, more precisely:
assert!((context_window as f64 * threshold) as usize > 0, "threshold produces zero limit");
```

decko (commented):
## Rust Specialist Review — round 2 (commit `85b0eda`)

All 4 findings from the previous round are addressed. 3 new findings on the fixes themselves:

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **High** | `estimate_payload` L77 | Double serialization failure overflows `usize` before `/4`: panics in debug, silently wraps to a tiny number in release — the conservative fallback is only effective when exactly one of the two fails |
| 2 | **Medium** | `is_over_threshold` L85 | `(total_input + total_output) as usize` is a narrowing cast on 32-bit targets (wasm32); once cumulative usage exceeds `u32::MAX` it wraps silently and the threshold never fires |
| 3 | **Low** | `is_over_threshold` L84 | `context_window = 1` passes the `assert!(> 0)` guard but produces `limit = 0` after f64 truncation — same always-true behavior as `context_window = 0` |

**Root pattern**: fixes #1 and #2 both use `as usize` casts without saturation. Switching the arithmetic in `estimate_payload` to `.saturating_add()` and keeping the threshold comparison in `u64` throughout `is_over_threshold` would close both at once.

